PR #8, opened 2021-02-14
🗺️  Add map wrapper component

### LINKED ISSUE #6: Add Google Maps wrapper component
## Description

Use [react-google-maps](https://www.npmjs.com/package/react-google-maps) to create a new component, `Map`, that is able to render locations and clusters on a Google Maps view. This component will *wrap* `react-google-maps`.

*This component will not interface with the API.* This is important because we'd like to separate out the API calls into the page component that will pass which clusters or locations to render to your `Map` component.

## Learning objectives
- Creating a "wrapper" component and why that might be better than using the third-party component directly
- Uncontrolled vs controlled components
- How to document component with JSDoc

## Todo

These points are a rough guideline. **Please** feel free to discuss with me and other devs about the best way to design this component!

- Make a draft PR for this, and push up changes often, so the rest of the team can see them
- Create a `components` directory in `src`
- Add a functional `Map` component
- Props
  - `onClusterSelect` callback that is called when a "cluster" is selected. Called with cluster id as the argument
  - `onLocationSelect` callback that is called when a "location" is selected. Called with location id as the argument
  - `onDeselect` callback that is called when the map is clicked, "deselecting" location or cluster. Maybe combine with `onClusterSelect` and `onLocationSelect`? Maybe combine all 3 together? Up to you.
  - `clusters` Array of objects `{id: xxx, lat: xxx, lng: xxx, radius: xxx, text: xxx}` displayed as colored circle on the map
  - `locations` Array of objects `{id: xxx, lat: xxx, lng: xxx}` displayed as pins on the map
  - props and callbacks for `center`, `zoom`, `bounds`. We want to make these fields **controlled**. In other words, the true `center` and `zoom` depend on the values of these props, not on internal state, and when these properties update (when a user zooms in, etc), call `onCenterChanged` or `onBoundsChanged` to let the parent component know. [Review what a controlled component is here.](https://reactjs.org/docs/forms.html) 
- Document your component with JSDoc and PropTypes!

## Useful links

- https://fallingfruit.org
- https://tomchentw.github.io/react-google-maps/#introduction
- https://petstore.swagger.io/?url=https://raw.githubusercontent.com/falling-fruit/api-docs/master/spec/v0.2.yaml
  - Although you will not be directly interacting with this API, you should get an idea of the response formats so you can make your component fit the data from the backend easily

laurenho025:
## Status: 
:rocket: Ready

## Description
Creates wrapper component for `google-map-react`. This component is controlled, receiving the map view state (center coordinates, zoom level, and bounds) as a prop, and propagating changes to the view state via the library's `onChange` event. Additional props include `locations` and `clusters` and their respective callbacks for when a location or cluster is clicked. 

Fixes #6 

## Todos
- [x] Documentation 

## Screenshots
Displays clusters for zoom level <= 12:
image

Displays locations for zoom level > 12:
image


src/App.js
```diff
@@ -13,7 +13,7 @@ const App = () => (
           <Types />
         </Route>
         <Route exact path="/">
-          <ExamplePage />
+          <HomePage />
         </Route>
         <Route>Not found</Route>
       </Switch>
```
Qrtn, line 19: For now, could you add a new route for your MapPage? Don't integrate into the tabs for now, still working out some details there.

src/components/Cluster.js
```diff
@@ -0,0 +1,46 @@
...
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 100%;
+  background: rgba(35, 74, 201, 0.214);
+  transform: translate(-50%, -50%);
+`
+
+const Cluster = (props) => {
+  const { onClick, lat, lng, count } = props
```
Qrtn, line 22: I think it's cleaner to unpack all props inside the function parameter, e.g.

`({ onClick, lat, lng count }) => {`

src/components/Cluster.js
```diff
@@ -0,0 +1,46 @@
...
+
+  const diameter = Math.min(
+    Math.max(
+      (Math.round(Math.log(count) / Math.log(10)) + 2) * 10,
+      MIN_CLUSTER_DIAMETER,
+    ),
+    MAX_CLUSTER_DIAMETER,
+  )
```
Qrtn, line 30: Can you move this to a helper function, like `calculateDiameter = (count) =>`? Might be easier to document that way.

src/components/Cluster.js
```diff
@@ -0,0 +1,46 @@
...
+  background: rgba(35, 74, 201, 0.214);
+  transform: translate(-50%, -50%);
+`
+
+const Cluster = (props) => {
+  const { onClick, lat, lng, count } = props
+
+  const diameter = Math.min(
+    Math.max(
+      (Math.round(Math.log(count) / Math.log(10)) + 2) * 10,
```
Qrtn, line 26: nit: `Math.log10(count)`

src/components/Location.js
```diff
@@ -0,0 +1,24 @@
...
+const LocationContainer = styled.button`
+  width: 10px;
+  height: 10px;
+  border-radius: 100%;
+  background: red;
+  transform: translate(-50%, -50%);
+`
+
+const Location = (props) => {
+  const { onClick, id } = props
```
Qrtn, line 15: Same thing with props unpacking as above.

src/pages/MapPage.js
```diff
@@ -0,0 +1,130 @@
...
+      zoom,
+    }))
+  }
+
+  const onDragEnd = ({ center }) => {
+    setView((prevState) => ({
+      ...prevState,
+      center: [center.lat, center.lng],
+    }))
+  }
```
Qrtn, line 97: It makes more sense to me to move the `onZoomAnimationEnd` and `onDragEnd` to the `Map` component, and instead expose a single `onViewChanged` callback. The `Map` component is meant to abstract away the lower-level details from whichever component uses it.

src/pages/MapPage.js
```diff
@@ -0,0 +1,130 @@
...
+const DEFAULT_CENTER_LAT = 40.1125785
+
+const DEFAULT_CENTER_LNG = -88.2287926
+
+const DEFAULT_ZOOM = 1
+
+const DEFAULT_VIEW_STATE = {
+  center: [DEFAULT_CENTER_LAT, DEFAULT_CENTER_LNG],
+  zoom: DEFAULT_ZOOM,
+  bounds: null,
```
Qrtn, line 60: Is `bounds` being updated here? If you expose `onViewChanged`, it would be the `Map` component's responsibility to update it.

`bounds` is necessary for the page to supply it when requesting clusters and locations.

src/components/Map.js
```diff
@@ -0,0 +1,62 @@
...
+            count={marker.count}
+          />
+        ),
+      )}
+    </GoogleMap>
+  )
+}
+
+Map.propTypes = {
+  bootstrapURLKeys: PropTypes.object.isRequired,
```
Qrtn, line 52: It may be best to expose only an `apiKey` prop here, rather than the entire `bootstrapURLKeys` object.

src/components/Map.js
```diff
@@ -0,0 +1,62 @@
...
+  view: PropTypes.object.isRequired,
+  markerData: PropTypes.arrayOf(PropTypes.object).isRequired,
+  showLocations: PropTypes.bool.isRequired,
```
Qrtn, line 55: It's a pretty interesting approach to take a list and a bool to represent one of [array of locations, array of clusters].

However, there are some potential downsides to this approach:

- Your location object and cluster object have a different schema. Cluster has a count property, and location has an id. In that sense, typing doesn't make sense here, because it's an array of two different types.
- Potentially, we may want to show both clusters and locations on the map. This is not possible with the current approach.
- Since you made parallel callbacks `onClusterClick`, `onLocationClick`, it would also make sense to make parallel array props for those. OTOH, you could make one callback, `onMarkerClick(details, isLocation)` but that doesn't seem clean to me.

In general it boils down to the fact that Cluster and Location are different types.

src/components/Map.js
```diff
@@ -0,0 +1,62 @@
...
+  } = props
+
+  return (
+    <GoogleMap
+      bootstrapURLKeys={bootstrapURLKeys}
+      center={view.center}
+      zoom={view.zoom}
+      onZoomAnimationEnd={onZoomAnimationEnd}
+      onDragEnd={onDragEnd}
+    >
```
Qrtn, line 27: (See below with the props. Consider just rendering both and leaving it up to the user of the component to clear out the opposite list.)

src/components/Cluster.js
```diff
@@ -0,0 +1,46 @@
...
+    <ClusterContainer diameter={diameter} onClick={() => onClick(lat, lng)}>
+      <p>{count}</p>
+    </ClusterContainer>
+  )
+}
+
+Cluster.propTypes = {
+  onClick: PropTypes.func.isRequired,
+  lat: PropTypes.number.isRequired,
+  lng: PropTypes.number.isRequired,
```
Qrtn, line 42: Is it necessary for a "Cluster" UI component to know about `lat` and `lng`? It gets rendered by Google Maps anyway in the correct location, right? `onClick` and `count` seem essential, OTOH.

src/components/Location.js
```diff
@@ -0,0 +1,24 @@
...
+  return <LocationContainer onClick={() => onClick(id)} />
+}
+
+Location.propTypes = {
+  onClick: PropTypes.func.isRequired,
+  id: PropTypes.number.isRequired,
```
Qrtn, line 21: Same with "Cluster": does "Location" have to know its `id` for anything?

src/components/Map.js
```diff
@@ -0,0 +1,62 @@
...
+      {markerData.map((marker, index) =>
+        showLocations ? (
+          <Location
+            key={marker.id}
+            onClick={onLocationClick}
```
Qrtn, line 32: In order to solve the issue of not knowing which location or cluster was clicked, I'd suggest that you make an anonymous function here that calls `onLocationClick` or `onClusterClick` with the entire location or cluster object, respectively.

In terms of data flow, it doesn't make sense to me that the UI components `Cluster` and `Location` should know about their individual properties (other than count, because that affects the diameter of the circle), since higher-level components can & should take care of that.

src/pages/MapPage.js
```diff
@@ -0,0 +1,130 @@
...
+    <MapContainer>
+      <Map
+        bootstrapURLKeys={{
+          key: `${process.env.REACT_APP_GOOGLE_MAPS_API_KEY}`,
+        }}
+        view={view}
+        markerData={markerData}
+        showLocations={showLocations}
+        onZoomAnimationEnd={onZoomAnimationEnd}
+        onDragEnd={onDragEnd}
```
Qrtn, line 122: See above about abstracting this away into the Map component.

Qrtn (changes requested):
Great work guys, you did an amazing job. I left a lot of comments because I really want you guys to think about component design and data flow in React. Let's feel free to continue this discussion at dev night.

src/components/Cluster.js
code-arman: lat, lng props are needed in order to render the components in the correct spot

src/components/Location.js
code-arman: Yes, we decided to pass the id to the location component because eventually we need to pass in the location id to get the location data

src/pages/MapPage.js
code-arman: ok, changed to onChangeHandler for now

src/App.js
```diff
@@ -12,10 +13,13 @@ const App = () => (
...
           <Types />
         </Route>
+        <Route exact path="/map">
+          <MapPage />
+        </Route>
         <Route exact path="/">
           <ExamplePage />
         </Route>
-        <Route>Not found</Route>
+        <Route>"Not found"</Route>
```
Qrtn, line 22: Just curious, was this a typo?

src/components/Cluster.js
```diff
@@ -0,0 +1,53 @@
...
+  transform: translate(-50%, -50%);
+`
+
+/**
+ * Component for a cluster displayed on the map.
+ * @param {function} onClick - The handler called when this cluster is clicked
+ * @param {number} count - The total number of locations within this cluster
+ */
+const Cluster = ({ onClick, count }) => {
+  const calculateDiameter = (count) =>
```
Qrtn, line 35: I'd recommend moving this helper function outside of the functional component, since it doesn't depend on component props or state. That way it's also easier to document.

You would want to leave it inside the component if it used count or onClick from the outside scope directly, rather than as a parameter (a closure)

Qrtn (approved):
Really beautiful work guys. Looking forward to seeing this getting hooked up to the API!

src/components/Location.js
```diff
@@ -0,0 +1,24 @@
...
+  border-radius: 100%;
+  background: red;
+  transform: translate(-50%, -50%);
+`
+
+/**
+ * Component for a location displayed on the map.
+ * @param {function} onClick - The handler called when this location is clicked
+ */
+const Location = ({ onClick }) => <LocationContainer onClick={onClick} />
```
Qrtn, line 18: Also, a styled-components tip here: you don't have to explicitly create a new `Location` component. This is the same as LocationContainer because buttons also have an onClick prop. So you could remove this line and rename the above to Location.

src/components/Cluster.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Component for a cluster displayed on the map.
+ * @param {function} onClick - The handler called when this cluster is clicked
+ * @param {number} count - The total number of locations within this cluster
+ */
+const Cluster = ({ onClick, count }) => (
+  <ClusterContainer diameter={calculateDiameter(count)} onClick={onClick}>
+    <p>{count}</p>
+  </ClusterContainer>
+)
```
Qrtn, line 50: See below with Location. You can do this calculation inside of the styled-component CSS block with templating, especially because you moved calculateDiameter outside.

Qrtn: Although this one is more debatable because you'd have to call calculateDiameter twice, for both width and height.

src/components/Map.js
```diff
@@ -0,0 +1,63 @@
...
+  <GoogleMap
+    bootstrapURLKeys={{ key: googleMapsAPIKey }}
+    center={view.center}
+    zoom={view.zoom}
+    onChange={onViewChange}
+  >
+    {clusters.map((cluster, index) => (
+      <Cluster
+        key={index}
+        onClick={() => onClusterClick(cluster.lat, cluster.lng)}
```
Qrtn, line 36: Ah, what I meant here is that you can pass the entire cluster object back, rather than just its latitude and longitude.

src/components/Map.js
```diff
@@ -0,0 +1,63 @@
...
+        count={cluster.count}
+        lat={cluster.lat}
+        lng={cluster.lng}
+      />
+    ))}
+    {locations.map((location) => (
+      <Location
+        key={location.id}
+        onClick={() => onLocationClick(location.id)}
```
Qrtn, line 45: Same thing here with location -- you might as well pass the entire location object back.

src/components/Cluster.js
laurenho025: Doesn't this need to be separate due to the child <p> inside of the cluster component?

Qrtn: Yes you're totally right! My bad.

Archna-1 (approved):
looks good! comments were really helpful when going through it :)

src/components/Location.js
Qrtn: I believe you can move the documentation to just above line 6, by the way.

PR #14, opened 2021-02-21
📦 Add Input, Tag

SirajChokshi:
## Status:

:rocket: Ready

## Description

Added basic UI elements:
- `Input`
- `TagList` and `Tag`
- Make the font-stack a theme variable
    - This is useful for HTML elements that override global fonts by default (e.g. `input`)

Fixes #<number>

## Todos

- [x] Finish `Tag`

## Screenshots
Screen Shot 2021-02-20 at 9 16 04 PM


src/components/ui/Input.js
```diff
@@ -0,0 +1,71 @@
...
+}
+
+const StyledInput = styled(Input)`
+  label {
+    color: ${({ theme }) => theme.secondaryText};
+    font-weight: bold;
+    line-height: 2;
+  }
+
+  & > div {
```
code-arman, line 33: do u really flexing on us with these crazy css selectors

src/components/ui/Input.js
```diff
@@ -0,0 +1,71 @@
+import styled from 'styled-components'
+
+const Input = ({ placeholder, onChange, onEnter, label, icon, className }) => {
+  const id = label?.toLowerCase().split(' ').join('-').concat(`-${className}`)
```
Qrtn, line 4: Are `name` and `id` necessary for the `input`? It doesn't seem like it gets used anywhere (no selection by id, and the input isn't being used by a form).

src/components/ui/Tag.js
```diff
@@ -0,0 +1,42 @@
+import styled from 'styled-components'
+
+const UnstyledTagList = ({ children, className }) => (
+  <ul className={className}>{children}</ul>
+)
+
+const TagList = styled(UnstyledTagList)`
+  padding: 0;
+  list-style: none;
+`
```
Qrtn, line 10: Here it seems better to style with

```
const TagList = styled.ul`
  padding: o;
  list-style: none;
`
```

You don't need the Unstyled intermediate tag.

src/components/ui/Tag.js
```diff
@@ -0,0 +1,42 @@
...
+  padding: 0 10px;
+  font-size: 12px;
+  font-weight: bold;
+  background: ${({ theme }) => theme.transparentOrange};
+  color: ${({ theme }) => theme.orange};
+
+  &:not(:last-child) {
+    margin-right: 6px;
+  }
+`
```
Qrtn, line 40: Same situation here. You can work off of `styled.li` directly, and change color and background color via `$(props) => props.color`, `$(props) => props.backgroundColor`

Qrtn: So you wouldn't need UnstyledTag at all.

src/pages/ExamplePage.js
```diff
@@ -21,6 +23,25 @@ const ExamplePage = () => (
...
+        <Input
+          label="Demo Form Input"
+          placeholder="'Sugar Maple'"
+          onChange={(e) => console.log(e.target.value)}
+        />
+        <br />
+        <TagList>
+          {['Rebecca', 'Jeffrey'].map((name) => (
+            <Tag key={name}>{name}</Tag>
+          ))}
```
Qrtn, line 42: Nice touch

Qrtn (changes requested):
The components look beautiful! 😍 Just a few suggestions.

src/components/ui/Input.js
Qrtn: Potentially you may want to spread remaining props into the HTML input, so that the user can pass their own id and name as they wish

SirajChokshi: Its an A11y feature. Name is only defined when there is a label so that when a user clicks said label the focus is put onto the relevant `<input />` element on the page. 

I will remove the `id` attribute though.

Qrtn: Oh gotcha, yes this is important due to the label!

Qrtn (approved):
Beautiful 😍

laurenho025 (approved):
lgtm

PR #18, opened 2021-02-22
🪟 Add adaptive layout

### LINKED ISSUE #17: Add adaptive layout
## Description

- Add breakpoints to website so that depending on window width, a mobile or desktop layout is rendered

Qrtn:
## Status:
:rocket: Ready

## Description

Adds adaptive layout.

- Adds DesktopBreakpoint and MobileBreakpoint components that conditionally render their children depending on if the device width is within a certain range
- Adds DesktopLayout and MobileLayout components
- Moves all components in `pages/` into the `components/` directory

Resolves #17

## Screenshots

image
image


src/components/desktop/DesktopLayout.js
```diff
@@ -0,0 +1,23 @@
...
+import MapPage from '../map/MapPage'
+import Header from './Header'
+
+const DesktopContainer = styled.div`
+  height: 100%;
+`
+
+const MapContainer = styled.div`
+  height: 100%;
+`
```
SirajChokshi, line 12: What do you think of writing a utility `Container` component here since we're using the same attributes to style both `DesktopContainer` and `MapContainer`?

src/components/map/MapPage.js
```diff
@@ -107,6 +101,7 @@ const MapPage = () => {
   }, [view.zoom])
 
   const onViewChange = ({ center, zoom, bounds }) => {
+    console.log('onViewChange called', { center, zoom, bounds })
```
SirajChokshi, line 104: Delete this `console.log`

src/components/MainPage.js
```diff
@@ -0,0 +1,16 @@
...
+
+const MainPage = () => (
+  <>
+    <DesktopBreakpoint>
+      <DesktopLayout />
+    </DesktopBreakpoint>
+    <MobileBreakpoint>
+      <MobileLayout />
+    </MobileBreakpoint>
+  </>
```
SirajChokshi, line 13: Thoughts on breaking up this child-rendering logic (from `Breakpoint.js`) into a chain of conditions? The current approach is functional, but the markup itself might be confusing since it does not imply only one layout is active at a time.

e.g. something like the following instead:
```javascript
import { useIsDesktop } from 'somewhere'

const MainPage = () => {
  const isDesktop = useIsDesktop()
  return isDesktop ? <DesktopLayout /> : <MobileLayout />
}
```

SirajChokshi (changes requested):
Great work 🔥 Just a few nits

src/components/map/MapPage.js
Qrtn: Thanks. I'm going to leave this one in for now to potentially help the devs debug with the map, because there's a really weird issue going on where the library we're using doesn't trigger `onViewChange` if the map isn't properly focused.

src/components/desktop/DesktopLayout.js
Qrtn: Things do get changed here: https://github.com/hack4impact-uiuc/falling-fruit/pull/20/files

The styling in this PR is actually a bit wrong because there's a scrollbar in the desktop layout (the map is too tall). But that gets fixed with flexbox in the pane PR.

src/components/MainPage.js
Qrtn: Great idea, that seems more clear to me.

SirajChokshi (approved):
lgmt 😇

laurenho025 (approved):
looks really good 😍

PR #20, opened 2021-02-22
🪟 Add resizable pane to desktop layout

### LINKED ISSUE #19: Add resizable pane to desktop layout
Add resizable pane to desktop layout. This will contain the search bar, filters, explore, options, etc.

Qrtn:
## Status:
:rocket: Ready

## Description

Adds resizable pane to desktop layout.

- Uses `react-split-pane` and `styled-components` to achieve this

Resolves #19 

## Screenshots

image
image


src/components/ui/ResizablePane.js
```diff
@@ -0,0 +1,18 @@
+import { Resizable } from 're-resizable'
+import styled from 'styled-components'
+
+const ResizablePane = styled(Resizable).attrs((props) => ({
+  // Mark defaultWidth as transient prop with $ so that it's not passed to the div
+  defaultSize: { width: props.$defaultWidth, height: '100%' },
+  enable: { right: true }, // enable resizing only to the right
```
laurenho025, line 7: Should `enable` also be passed as a prop? In case we happen to reuse ResizablePane somewhere.

src/components/desktop/DesktopLayout.js
```diff
@@ -0,0 +1,37 @@
...
+  right: 0;
+  height: 100%;
+  width: calc(100% - 310px);
+`
+
+const DesktopLayout = () => (
+  <DesktopContainer>
+    <Header />
+    <ContentContainer>
+      <ResizablePane minWidth={310} maxWidth="21.5vw" $defaultWidth="21.5vw" />
```
laurenho025, line 29: Small thing, but maybe consider extracting these props into consts or a single const object? Not sure if it would be cleaner or not...

src/components/ui/ResizablePane.js
Qrtn: Ah yeah this is a good point! So I would remove line 7 and styled will automatically pass it through.

src/components/desktop/DesktopLayout.js
Qrtn: I had to switch the side panel from being an "overlay" to an actual pane, so this code got updated. Did end up using constants for this because I had to kind of hack in the max size to work with the library I'm using. You can see in the latest commit.

src/App.js
```diff
@@ -1,28 +1,31 @@
...
       <Switch>
-        <Route exact path="/types">
-          <Types />
+        <Route path="/desktop">
+          <DesktopLayout />
         </Route>
-        <Route exact path="/map">
-          <MapPage />
+        <Route path="/mobile">
+          <MobileLayout />
```
code-arman, line 18: This is here just for testing stuff right?

src/components/desktop/Header.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components'
+
+// Might become a navbar, so use <nav> for this
+const DesktopHeader = styled.header`
```
code-arman, line 4: Are you suggesting you wanted to use styled.nav instead of styled.header?

code-arman (approved):
Looks pretty fire to me

src/App.js
Qrtn: Yes! Although it might be nice to have specific URLs targeting mobile or desktop layouts.

src/components/desktop/Header.js
Qrtn: Yes, but not totally sure if this header is a nav yet. Waiting on
@SirajChokshi's content inside the bar.

vasuchal (approved):
Great job

PR #33, opened 2021-03-04
🕵️ Entry Details

Archna-1:
## Status: 
:rocket: Ready

## Description
Show entry details page and entry nav when a location is clicked on the map. Adds routing in side pane to display entry details wrapper on desktop view. 

Fixes #33  

## Screenshots

Entry details page:
Screen Shot 2021-03-10 at 6 36 18 PM

Raised & flat icon buttons:
image


package.json
```diff
@@ -7,6 +7,7 @@
     "@reach/window-size": "^0.13.2",
     "@styled-icons/boxicons-regular": "^10.23.0",
     "@styled-icons/boxicons-solid": "^10.23.0",
+    "@styled-icons/fa-solid": "^10.26.0",
```
Qrtn, line 10: We're using boxicons for all our icons, so try to find the equivalent in box icons (either regular or solid) and remove font awesome. Thanks!

src/components/entry/EntryNav.js
```diff
@@ -1,7 +1,70 @@
...
+          size={50}
+          raised={false}
+          icon={<Pencil size="20" color={theme.secondaryText} />}
+          onClick={onEditEntryDetailsButtonClick}
+          label={'edit-entry-details'}
+        />
+        <IconButton
+          size={50}
+          raised={false}
+          icon={<Map size="20" color={theme.secondaryText} />}
```
Qrtn, line 61: I think it'd be a good idea to style the icon itself (you can select with `svg {` in StyledIconButton) to be height 50% of container. This way we don't have to pass in an explicit size for the icon, just a size in IconButton. You can ref Siraj's comments in the design channel:

> Hey guys — for resizable components (like IconButton) we should have size be a prop that takes a number.  Centered child icons can have a width and height of 50% of the size. Since these components are used frequently at many sizes throughout the interface its beneficial to be able to modify them for each use case.

I believe @SirajChokshi mentioned to only set one of the dimensions of the icon to 50%, just in case the icon is rectangular.

src/components/ui/IconButton.js
```diff
@@ -0,0 +1,61 @@
...
+const RaisedIconButtonContainer = styled.span`
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  border-radius: 50%;
+  width: ${({ size }) => size}px;
+  height: ${({ size }) => size}px;
+  border: 2px solid;
+  border-color: orange;
+`
```
Qrtn, line 34: Would be nice to look into ways to achieve multiple border without an extra container: https://css-tricks.com/snippets/css/multiple-borders/

Don't worry about this now though!

src/components/entry/EntryDetails.js
```diff
@@ -7,12 +44,38 @@ const EntryDetails = () => {
...
+      />
+
+      <TextContainer>
+        <EntryHeader>
+          <h3>American Tulip Tree</h3>
+          <Tag>Invasive</Tag>
+          <Tag>Private, But Overhanging</Tag>
+        </EntryHeader>
+
+        <>
```
Qrtn, line 59: Remove these unnecessary fragments

src/components/entry/EntryDetails.js
```diff
@@ -1,4 +1,41 @@
...
+const EntryHeader = styled.div`
+  line-height: 12px;
+`
+
+/* Row gap nor line-height is working here to make the distance between all elements 14px
+const DescriptionContainer = styled.div`
+  grid-template-rows: 100%
+  row-gap: 14px;
+`
+*/
```
Qrtn, line 32: Talk to @SirajChokshi or @laurenho025 about the best way to achieve this. One way is to use a CSS selector to select all immediate children: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors. This might not be necessary though

src/components/entry/EntryNav.js
```diff
@@ -1,7 +1,70 @@
...
+
+  const onViewEntryOnMapButtonClick = () => {
+    // TODO: view entry on map callback
+    console.log('View entry on map clicked')
+  }
+
+  return (
+    <EntryNavContainer>
+      <EntryNavTextContainer>
+        <ArrowBack size="20" color={theme.secondaryText} />
```
Qrtn, line 47: Also tip on implementing the back button. Since we have routing you should be able to just `history.pop`. That'll take you back to the right view, regardless if it was map or list.

src/components/entry/EntryNav.js
```diff
@@ -1,7 +1,70 @@
...
+
+const EntryNavIconsContainer = styled.div`
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  width: 110px;
+`
+
+const EntryNav = () => {
+  const onEditEntryDetailsButtonClick = () => {
```
Qrtn, line 34: Maybe `onEditButtonClick` and `onMapButtonClick`

src/components/entry/EntryNav.js
```diff
@@ -1,7 +1,70 @@
...
+import { Pencil } from '@styled-icons/boxicons-solid'
+import { MapMarkerAlt as Map } from '@styled-icons/fa-solid'
+import styled from 'styled-components'
+
+import { theme } from '../ui/GlobalStyle'
+import IconButton from '../ui/IconButton'
+
+const EntryNavContainer = styled.div`
+  display: flex;
+  flex-direction: row;
```
Qrtn, line 11: You can omit the `flex-direction: row` in all the containers because it's the default.

src/components/entry/EntryNav.js
```diff
@@ -1,7 +1,70 @@
...
+
+const EntryNavTextContainer = styled.div`
+  display: flex;
+  align-items: center;
+  flex-direction: row;
+`
+
+const EntryNavText = styled.p`
+  margin-left: 10px;
+`
```
Qrtn, line 24: Maybe remove this styled component and put the margin on the svg. With minor/less semantic styling in styled components, we want to avoid creating new containers by using selectors on the main component instead.

Qrtn: So in this case in EntryNavTextContainer

src/components/entry/EntryNav.js
laurenho025: whoops I completely forgot about this, thanks!

src/components/entry/EntryDetails.js
```diff
@@ -1,4 +1,82 @@
...
+  .then(function (response) {
+    console.log(response);
+  })
+  .catch(function (error) {
+    console.log(error);
+  })
+  .then(function () {
+    // always executed
+  });  
+*/
```
Qrtn, line 24: Make sure to remove this

src/components/entry/EntryDetails.js
```diff
@@ -1,18 +1,140 @@
...
+      </TextContainer>
+    </div>
+  ) : (
+    <div>
+      <br />
       <br />
       <br />
       <br />
       <br />
       <br />
```
Qrtn, line 136: You can remove these `br`s I put in lol

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,239 @@
...
+          <IndividualResourceContainer>
+            <img
+              src={FruitipediaLogo}
+              height="20px"
+              width="25px"
+              alt="Fruitipedia logo"
+            />
+            <a href={locationTypeData.fruitipedia_url}>Fruitipedia</a>
+          </IndividualResourceContainer>
+        )}
```
Qrtn, line 230: Noticed you have a lot of code repetition here, from lines 170-230. Could you try to convert this into a loop?

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,253 @@
...
+  margin-bottom: 0px;
+  color: #333333;
+`
+
+const ScientificName = styled.small`
+  font-style: italic;
+`
+
+// Wraps the plant name and scientific name, as well as an icon button for desktop
+const HeaderContainer = styled.div`
```
rxun, line 62: it may be possible to convert a few divs you have to a more semantic sections format (https://html.spec.whatwg.org/multipage/sections.html) -- e.g. HeaderContainer maybe could be a header, EntryDetailsContent could be an article, DescriptionContainer could potentially be a section, etc. i think that styling should be unaffected? i think i'll leave this more to your discretion, however, because:

> "The section element is not a generic container element. When an element is needed only for styling purposes or as a convenience for scripting, authors are encouraged to use the div element instead. A general rule is that the section element is appropriate only if the element's contents would be listed explicitly in the document's outline."

and mostly, your containers have to do with styling so they look good!

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,253 @@
...
+import { theme } from '../ui/GlobalStyle'
+import IconButton from '../ui/IconButton'
+import { Tag } from '../ui/Tag'
+import EatTheWeedsLogo from './EatTheWeeds.svg'
+import ForagingTexasLogo from './ForagingTexas.png'
+import FruitipediaLogo from './Fruitipedia.png'
+import UrbanMushroomsLogo from './UrbanMushrooms.png'
+import USDALogo from './USDA.svg'
+
+const ACCESS_TYPE = {
```
rxun, line 19: i think it'd be useful to extract this into its own utils file potentially, since there are other enums also associated with entry details (Fruiting Status, Quality Rating, and Yield Rating) that can be used elsewhere in the future. 

the list of logos/urls could also potentially use its own utils file like ymca did here: https://github.com/hack4impact-uiuc/ymca/blob/master/client/src/utils/determineStockPhoto.js

Qrtn (approved):
LGTM!! Great job Archna and Lauren! 😻

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,204 @@
...
+}
+
+const formatISOString = (dateString) =>
+  new Date(dateString).toLocaleDateString(undefined, {
+    year: 'numeric',
+    month: 'long',
+    day: 'numeric',
+  })
+
+// TODO: Reduce number of styled components by using selectors in the container
```
code-arman, line 28: r u gonna do this in the PR?

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,204 @@
...
-      <br />
-      <br />
-      <br />
-      <p>EntryDetails for id: {id}</p>
-    </div>
+  const [locationData, setLocationData] = useState()
+  const [locationTypeData, setLocationTypeData] = useState()
+
+  useEffect(() => {
+    async function fetchEntryDetails() {
```
code-arman, line 116: can this be an arrow function?

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,204 @@
...
+            {title}
+          </ResourceLink>
+        </IndividualResourceContainer>
+      ),
+  )
+
+  return locationData && locationTypeData ? (
+    <EntryDetailsPageContainer isDesktop={isDesktop}>
+      {locationData.photos.length > 0 && (
+        <ImageContainer src={locationData.photos[0].photo.original} alt="" />
```
code-arman, line 150: where the alt text at. #a11y

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,204 @@
...
+              raised={false}
+              icon={<Map color={theme.secondaryText} />}
+              onClick={handleMapButtonClick}
+              label="add location"
+            />
+          )}
+        </HeaderContainer>
+        <TagList>
+          {locationData.access && <Tag>{ACCESS_TYPE[locationData.access]}</Tag>}
+          {/* TODO: Put tag colors in theme/use constants somehow */}
```
code-arman, line 171: r u gonna do this in this pr?

src/components/entry/icons/EatTheWeeds.svg
```diff
@@ -0,0 +1,3 @@
...
...a8Thp1mN++S5Sa4lx6YEkCesMiwHGlCihPIB+2kPJUZK4ZEnzLqu9Bz6ljMEr/WH7IdcezAKGg48cZefEHqqVmIquE6mEE/tOs2dyPydnnqOfdznzt9e4tzaPdnkX6NBTqbXmLxff4Hd/fYVapUGoKtxbnSeJq6RpShIlxEHN02YxFFlBpEOf1x9cf48TM88wUd3Gd7/0A/++mUx6b5ioPsLL3/wJM5OHmVu7wq/ffI3JPQkyzwqq1dgHd9fWxhRb6hMkUZVG0mTf9gNOlT5YpGMPtBJUEcIgnW0IgUJxY+FffDh7jh1b9jLd3E0c1pi9d5FLCxcIVMDuyf3eYOZac3T6HaQzGec0YRChdUhZGo7tOc3xvV9FycCD+fTWeT65830XA61Cv5K1dJmJZMrrJLAWqSR5lvPqG69w8sgpYlWn1V/g3c/eQlrJU7tP0+20qMVjvPP560gpCSoVtHtAQi/tUJjSoxNOjQjqcYMbi7Ok/dSbTndjjaXuAvVojK4Yerr/Y7NhEHJzZZ6P597nie3HmVu5xUe3P0BkitsPr1KJQs5e+j23lq4RyIgyLxGnfrTLhlGF8S2T2LwkjCN0RWMKS6hDWmvLrG10yIeWei1mesc2olrE6vIquuqAS3rd1AOev9ri2LMHSHRMq92i1Vpl0Bsy1thCLanRz9sYa7g92+axPTW0jgLvx0ud+5jSQk8RJ4psWHrzF0IhhcKanMxkrKyvUA+qpHlKnaovPp69zZozLAeUDH2WOAEH2tDutunnPeJIo6UrNBahNNqVPxUoKAVRpN1rgiBE2MLP51KqKEpG2pK4n7CCqDLSiFRqE+SoGiqhfWoKD2g0TvtMClDamYtLW1cxBb7QCSU8AGeHroQ6cTlHMi6ocyQ5mtile5aVvoy6d9YlOqOyqrQajTHWFyYxkoP3ATywzdYDhjLLkEVpRhNrSVzRVGLlFe06OyG5IE7Nrk8xNIxmHd15Vm6yvVmpjfHBR5XbjoBs1nnhycYv1C3CMebw+9WWufNcg3JQXQyvYuF14OkS+BW5GcrMoJSkcIoFBhuZjz+i//+PHSNYjkm8OckgRIeKIIndSUaOTiHGMBwahoMREw6Q2zY33O27ZyHwL/wWue/Omv02KRcZgtCx5kDaEd3+xPK/1sVwKVaWljwd8G8Z8jybYVC5EwAAAABJRU5ErkJggg=="/>
```
code-arman, line 2: this is massive

code-arman (approved):
lgtm, but use better emoji. like maybe:  ℹ️  or 🕵️  or something else. :)

src/components/entry/EntryDetails.js
Qrtn: No, this will be in a separate PR!

src/components/entry/EntryDetails.js
Qrtn: They are the same in this case. I was thinking about how we should standardize arrow vs. standard functions throughout the repo (when to use which one). Would like to get everyone's thoughts on this. For now, it should be ok.

src/components/entry/EntryDetails.js
Qrtn: Yes, it's best to add an alt text here.

src/components/entry/EntryDetails.js
Qrtn: We'll do this in a later PR.

src/components/entry/EntryDetails.js
laurenho025: Doing it now

src/components/entry/icons/EatTheWeeds.svg
Qrtn: Yes, this should be the original image, not converted to an SVG. But we can do this in a later PR. Please add a TODO somewhere in resources.js!

src/components/entry/EntryDetails.js
Qrtn: We can do this in a later PR though. Can you add a TODO here?

src/components/entry/icons/EatTheWeeds.svg
code-arman: ^ @laurenho025

src/components/desktop/MainPane.js
```diff
@@ -1,9 +1,9 @@
 import SearchOverlay from './SearchOverlay'
 
-const SidePane = () => (
+const MainPane = () => (
   <>
```
SirajChokshi, line 4: Do we need this fragment here? My understanding is that this component will probably have more children later (otherwise it does not need to exist).

Maybe we could add a `// TODO!  ...` above this?

src/components/desktop/SidePaneSwitch.js
```diff
@@ -0,0 +1,17 @@
+import { Route, Switch } from 'react-router-dom'
+
+import EntryPane from './EntryPane'
+import MainPane from './MainPane'
+
+const SidePaneSwitch = () => (
+  <Switch>
+    <Route path="/entry/:id">
```
SirajChokshi, line 8: Can we add the `exact` property here for clarity?

src/components/desktop/SidePaneSwitch.js
```diff
@@ -0,0 +1,17 @@
...
+      <EntryPane />
+    </Route>
+    <Route>
```
SirajChokshi, line 11: It seems the main page catches all other routes (acts as a default case when previous routes fall through). Could we enforce an exact blank route (`/`) here and instead default to a redirect to the blank route?

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,207 @@
...
+
+const ACCESS_TYPE = {
+  0: "On lister's property",
+  1: 'Received permission from owner',
+  2: 'Public property',
+  3: 'Private but overhanging',
+  4: 'Private property',
+}
+
+const formatISOString = (dateString) =>
```
SirajChokshi, line 21: A comment here describing how to interface with this function (return types on error, etc) would be useful

src/components/entry/EntryDetails.js
```diff
@@ -1,19 +1,207 @@
...
+    }
+    fetchEntryDetails()
+  }, [id])
+
+  const handleMapButtonClick = () => {
+    // TODO: handle map button click
+    console.log('Map Button Clicked')
+  }
+
+  const resources = RESOURCES.map(
```
SirajChokshi, line 131: Great work on this :)

src/components/entry/resources.js
```diff
@@ -0,0 +1,49 @@
+import EatTheWeedsLogo from './icons/EatTheWeeds.svg'
+import ForagingTexasLogo from './icons/ForagingTexas.png'
+import FruitipediaLogo from './icons/Fruitipedia.png'
+import UrbanMushroomsLogo from './icons/UrbanMushrooms.png'
+import USDALogo from './icons/USDA.svg'
+import WikipediaLogo from './icons/Wikipedia.svg'
+
+/* TODO: Convert logos to original images instead of SVGs */
+const RESOURCES = [
```
SirajChokshi, line 9: Include a jsdoc/equivalent comment to describe this constant

src/components/desktop/MainPane.js
Qrtn: Yes, it'll have more children later! TODO would be nice.

src/components/desktop/SidePaneSwitch.js
Qrtn: Yes, good idea.

src/components/desktop/SidePaneSwitch.js
Qrtn: This route should catch /map and /list without redirecting. You're right, we may want to explicitly have a route for each of those, with MainPane.

PR #36, opened 2021-03-14
✅ Checkbox component

vasuchal:
## Description
Checkbox component that wraps CustomCheckbox from React UI and has three states: mixed, true, false

Fixes #<number>

## Screenshots
Screen Shot 2021-03-14 at 3 35 59 AM


src/components/ui/Checkbox.js
```diff
@@ -0,0 +1,79 @@
...
+import { Check } from '@styled-icons/boxicons-regular'
+import React from 'react'
+import styled from 'styled-components'
+
+const Checkbox = React.forwardRef((props, ref) => (
+  <div className={props.className}>
+    <CustomCheckboxContainer>
+      <CustomCheckboxInput
+        checked={props.checked}
+        onChange={props.onChange}
```
Qrtn, line 13: `{ className, checked, onChange, ...props }`

src/components/ui/Checkbox.js
```diff
@@ -0,0 +1,79 @@
...
+    top: 20%;
+    left: 20%;
+    width: 60%;
+    height: 60%;
+    transition: transform 200ms ease-out, background 200ms ease-out;
+    zindex: 1;
+    background: ${(props) => {
+      if (props.checked === true) {
+        return ({ theme }) => theme.orange
+      } else if (props.checked === false) {
```
SirajChokshi, line 65: Since there are only two outputs to this function can we check if `props.checked === false` initially and then return our `else` condition from there?

src/components/ui/Checkbox.js
```diff
@@ -0,0 +1,79 @@
...
+      <CustomCheckboxInput
+        checked={props.checked}
+        onChange={props.onChange}
+        ref={ref}
+        {...props}
+      />
+      {props.checked === true ? (
+        <StyledCheck size="21" />
+      ) : (
+        <span aria-hidden></span>
```
SirajChokshi: A self-closing tag would make it clear we don't intend to put children here.

src/components/ui/Checkbox.js
```diff
@@ -0,0 +1,79 @@
...
+    background: ${(props) => {
+      if (props.checked === true) {
+        return ({ theme }) => theme.orange
+      } else if (props.checked === false) {
+        return ({ theme }) => theme.transparentOrange
+      } else {
+        return ({ theme }) => theme.orange
+      }
+    }};
+    transform: ${(props) => {
```
SirajChokshi, line 71: This style doesn't seem to do anything, can we delete it?

SirajChokshi (commented):
Good work, few nits

code-arman:
☑️ -- use an emoji like this at the front :)

Qrtn:
> ☑️ -- use an emoji like this at the front :)

Mm yeah +1 😛

Qrtn (approved):
LGTM, great work! 🙌

Qrtn:
Also Vasu, please fill in the PR description when you get a chance.

src/components/ui/Checkbox.js
vasuchal: only the mixed state if is there now and we need it because it gives the background color of the square

SirajChokshi: I'm referring to the transform. If the output is
```js
if (props.checked === 'mixed') {
  return `scaleX(1) scaleY(1)`
} else {
  return `scaleX(1) scaleY(1)`
}
```
and scaling to 1 in both X and Y dimensions results in the same output as doing nothing. What does this property do in the case of 'mixed'?

vasuchal: ohh ok I see yeah there's no point of it being there

SirajChokshi (approved):
Great work on this :)

PR #39, opened 2021-03-14
🔳 Add border to desktop search overlay

Qrtn:
## Status:
:rocket: Ready

## Description

Adds a bottom border to the search overlay in desktop. It only shows when the popover with suggestions is open, which works for our use case. Will need to be updated when filtering is included here.

Also added this Babel plugin (to use it just `import styled from 'styled-components/macro'`): https://github.com/styled-components/babel-plugin-styled-components. We get nice things like this (note the class names):

image

## Screenshots

image

### DIFF
--- a/src/components/desktop/SearchOverlay.js
+++ b/src/components/desktop/SearchOverlay.js
@@ -9,7 +9,7 @@ const SearchOverlay = styled(Search)`
   }
 
   [data-reach-combobox-popover] {
-    // TODO: this is tricky, but make highlighting contained wihtin the border radius
+    // TODO: this is tricky, but make highlighting contained within the border radius
     border-bottom-left-radius: 30px;
     border-bottom-right-radius: 30px;
     padding-bottom: 10px;
--- a/src/components/desktop/SearchOverlay.js
+++ b/src/components/desktop/SearchOverlay.js
@@ -1,0 +1,21 @@
+import styled from 'styled-components'
+
+import Search from '../search/Search'
+
+const SearchOverlay = styled(Search)`
+  & > div:first-child {
+    // ComboboxInput
+    padding: 10px 10px 0 10px;
+  }
+
+  [data-reach-combobox-popover] {
+    // TODO: this is tricky, but make highlighting contained wihtin the border radius
+    border-bottom-left-radius: 30px;
+    border-bottom-right-radius: 30px;
+    padding-bottom: 10px;
+    // TODO: siraj fix border radius ^ and box shadow here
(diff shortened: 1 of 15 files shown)


src/components/map/MapPage.js
```diff
@@ -46,7 +46,7 @@ const MapPage = () => {
 
   useEffect(() => {
     async function fetchClusterAndLocationData() {
-      if (view.bounds) {
+      if (view.bounds?.ne.lat) {
```
SirajChokshi, line 49: Weird edge case, but in South America or Africa this value could be zero and thereby false-y.

src/components/desktop/SearchOverlay.js
```diff
@@ -0,0 +1,21 @@
...
+import Search from '../search/Search'
+
+const SearchOverlay = styled(Search)`
+  & > div:first-child {
+    // ComboboxInput
+    padding: 10px 10px 0 10px;
+  }
+
+  [data-reach-combobox-popover] {
+    // TODO: this is tricky, but make highlighting contained wihtin the border radius
```
SirajChokshi, line 12: typo -- sorry for being a pedant 🥴

src/components/desktop/SidePane.js
```diff
@@ -1,15 +1,12 @@
...
+import styled from 'styled-components/macro'
 
-import Search from '../search/Search'
+import SearchOverlay from './SearchOverlay'
 
-const PaneContainer = styled.div`
-  padding: 20px 10px;
-  padding-right: 0px;
-`
+const PaneContainer = styled.div``
```
SirajChokshi, line 5: Since `PaneContainer` is has an empty ruleset could we make a div instead of defining a new styled-component?

SirajChokshi (commented):
Great work, few very minor nits

laurenho025 (approved):
looks good 🤩

src/components/map/MapPage.js
Qrtn: Good point, better to be safe here.

src/components/desktop/SearchOverlay.js
Qrtn: Haha no i appreciate it!!

src/components/desktop/SidePane.js
Qrtn: Good point - changed to a fragment. Left that there bc I thought it might need styles later but yeah seems better to take it out for now

SirajChokshi (approved):
Nice work!

PR #69, opened 2021-04-03
🧑🏼‍🦽 Form progress bar

### LINKED ISSUE #58: Form progress bar
## Description

Create progress bar showing the steps in the form:

image

`ui/ProgressBar.js`

Branch name: `form-progress`

## Todo

- Create the component in ui
- Break down the whole progress bar component into multiple components for easier styling and separation of concerns
- Decide on what makes sense to take as props for this component
  - currentStep
  - onChange
  - labels: an array of N labels. This also determines the number of steps (N)
- Styling
  - Track, circles, labels under the track
- Integrate props
- Test

## Useful links

- https://medium.com/@beyondborders/building-a-responsive-and-dynamic-progress-bar-5d52c1f66c24
- Siraj

vasuchal:
## Description
Progress Bar Component for Form. This bar in the form allows users to see what part they are at.

Fixes #58 

## Screenshots
Screen Shot 2021-04-07 at 8 32 16 PM
Screen Shot 2021-04-07 at 8 32 28 PM
Screen Shot 2021-04-07 at 8 32 36 PM


src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,46 @@
...
+    right: 36px;
+    height: 100%;
+    position: absolute;
+    align-self: center;
+    content: '';
+    border-top: 2.5px solid black;
+    width: calc(100% - 34px);
+  }
+`
+StyledProgressBarStep.displayName = 'ProgressBarStep'
```
Qrtn, line 44: Can you do the same rearrangement with the styled component as we did in ProgressBar?

src/components/ui/ProgressBar.js
```diff
@@ -0,0 +1,34 @@
...
+import styled from 'styled-components/macro'
+
+import ProgressBarStep from '../ui/ProgressBarStep'
+
+const StyledProgressBar = styled.div`
+  display: flex;
+  justify-content: space-between;
+`
+const getStepStatus = (index, currentStep) => {
+  const refactoredIndex = index + 1
```
Qrtn: Can you explain why this is necessary?

src/components/ui/ProgressBar.js
```diff
@@ -0,0 +1,35 @@
...
+import styled from 'styled-components/macro'
+
+import ProgressBarStep from '../ui/ProgressBarStep'
+
+const StyledProgressBar = styled.div`
+  display: flex;
+  justify-content: space-between;
+`
+const getStepStatus = (index, currentStep) => {
+  const refactoredIndex = index
```
Qrtn, line 11: Is this line still necessary?

src/components/ui/ProgressBar.js
```diff
@@ -0,0 +1,35 @@
...
+  if (refactoredIndex === currentStep) {
+    return 'active'
+  } else if (refactoredIndex < currentStep) {
+    return 'complete'
+  } else {
+    return 'incomplete'
+  }
+}
+
+const ProgressBar = ({ labels, className, currentStep, onChange }) => (
```
Qrtn, line 21: Can you replace all mentions of currentStep with step? Don't think current is necessary anymore.

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,75 @@
+import { Check } from '@styled-icons/boxicons-regular'
+import React from 'react'
+import styled from 'styled-components/macro'
+
+import CircleIcon from '../ui/CircleIcon'
+import { theme } from './GlobalStyle'
+
+const PrimaryText = styled.div`
+  font-size: 15px;
+  color: ${({ color }) => color};
```
Qrtn, line 10: Make color a transient prop as well

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,75 @@
+import { Check } from '@styled-icons/boxicons-regular'
+import React from 'react'
+import styled from 'styled-components/macro'
+
+import CircleIcon from '../ui/CircleIcon'
+import { theme } from './GlobalStyle'
+
+const PrimaryText = styled.div`
```
Qrtn, line 8: Can you check if you need this styled component at all? You might be able to just move these styles into StyledCircleIcon and get rid of the wrapping div around the step number text. If that doesn't work, you can just style using the selector div {}. Creating a new one seems a bit too much.

Then you can just switch color based off status which is cleaner

Qrtn: You can also access the theme inside StyledCircleIcon

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,75 @@
+import { Check } from '@styled-icons/boxicons-regular'
+import React from 'react'
+import styled from 'styled-components/macro'
```
Qrtn, line 3: There's been a couple usages of this in the past where it's been hard to avoid, but importing the theme is not good practice. You should access it inside the styled component. I see a couple usages here in PrimaryText where you can avoid this

src/components/ui/ProgressBarStep.js
Qrtn: Or even style all text related stuff in StyledProgressBarStep. Then you might only have to do things like font-size once.

src/components/ui/ProgressBarStep.js
Qrtn: Importing the theme is not good practice because then it's not swappable

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,75 @@
...
+`
+
+const ProgressBarStep = ({ label, stepNumber, status, onClick }) => (
+  <StyledProgressBarStep $status={status}>
+    <StyledCircleIcon
+      backgroundColor={
+        status === 'incomplete'
+          ? theme.secondaryBackground
+          : theme.transparentOrange
+      }
```
Qrtn, line 53: Maybe override this via StyledCircleIcon actually

Qrtn (changes requested):
Great work Vasu, this was a tough task. I made quite a few styling comments because I want the whole team to think critically about the best way to use styled-components. The library is incredibly powerful, but we should still take care. So hopefully we can discuss in this PR.

@SirajChokshi would appreciate thoughts as well, in addition to Vasu's

src/components/ui/ProgressBar.js
Qrtn: Made duplicate comment about this LOL, previous comment was for the outdated version where `index + 1`

Qrtn: But yeah you can remove this

src/components/ui/ProgressBarStep.js
vasuchal: Instead of having color as a transient prop, is there a reason we didn't do color: ${({ theme }) => theme.primaryText}?

Qrtn: Your color is dynamic, so now you can just switch it based off status. I made this comment before the ones above

src/components/ui/ProgressBarStep.js
Qrtn: The point is to not have to import theme--you can access the theme within the CSS

package.json
```diff
@@ -9,6 +9,7 @@
     "@reach/accordion": "^0.13.2",
     "@reach/checkbox": "^0.13.2",
     "@reach/combobox": "^0.13.2",
+    "@reach/dialog": "^0.15.0",
```
Qrtn, line 12: Is this for a separate PR?

src/components/mobile/SettingsPage.js
```diff
@@ -13,63 +15,73 @@ import {
...
+          </AccordionPanel>
+        </AccordionItem>
+      </SettingsAccordion>
+      <Slider
+        style={{ margin: '0 40px' }}
+        labels={['Label 1', null, 'Label 3', 'Label 4', 'Label 5']}
+        steps={5}
+      />
+      <br></br>
+      <br></br>
```
Qrtn, line 77: Can be written as `<br />`

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,67 @@
...
+`
+const StyledCheck = styled(Check)`
+  color: ${({ theme }) => theme.orange};
+`
+
+const ProgressBarStep = ({ label, stepNumber, status, onClick }) => (
+  <StyledProgressBarStep $status={status}>
+    <StyledCircleIcon $status={status} onClick={onClick}>
+      {status === 'complete' ? (
+        <StyledCheck></StyledCheck>
```
Qrtn, line 58: `<StyledCheck />`

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,67 @@
...
+      ) : (
+        <div> {stepNumber} </div>
```
Qrtn, line 60: Do you still need this div?

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,67 @@
...
+      )}
+    </StyledCircleIcon>
+    <div> {label} </div>
```
Qrtn, line 63: Same here - is this div necessary? If yes, then you can def keep, but try taking this one and the one above out individually to see if they're necessary.

Qrtn: Remove extra spaces around braces

src/components/ui/ProgressBar.js
```diff
@@ -0,0 +1,34 @@
+import { React } from 'react'
```
Qrtn, line 1: You can remove this import

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,67 @@
+import { Check } from '@styled-icons/boxicons-regular'
+import React from 'react'
```
Qrtn, line 2: You can remove this import as well.

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,67 @@
...
+
+  // TODO: fix specificity of ::before CSS
+
+  &::before {
+    border-color: ${({ $status, theme }) =>
+      $status === 'incomplete' ? theme.secondaryText : theme.orange} !important;
+  }
+`
+const StyledCheck = styled(Check)`
+  color: ${({ theme }) => theme.orange};
```
Qrtn, line 51: You can style this with an `svg {}` selector in StyledCircleIcon.

Qrtn: This way you can avoid creating a separate styled component

src/components/ui/ProgressBarStep.js
```diff
@@ -0,0 +1,67 @@
...
+`
+
+const ProgressBarStep = ({ label, stepNumber, status, onClick }) => (
```
Qrtn, line 54: I think it's actually more semantic to change `stepNumber` to the `children` prop. Make sure to pass that in correctly in ProgressBar.js (not just `children={1}`).

src/components/ui/ProgressBarStep.js
vasuchal: yep, I tried before without the divs and it messed up styling

package.json
vasuchal: yeah i added it to my other branch but it added it here, should I delete it?

Qrtn: Yes.

code-arman (approved):
LGTM

PR #70, opened 2021-04-04
🌳 Add form type selection

### LINKED ISSUE #59: Form type selection
## Description

Refactor the existing tree select component to handle displaying the currently selected types, and so that it's not tied to context (tied to props). Then, there will be two instances of the tree select: one in filter, one in the new location form. Take a prop that shows whether or not to display tags.

image

Branch name: `form-type-selection`

## Todo

- Refactor out existing component
- Style tags
- Take a prop that shows whether or not to display tags
- When the component is used for selection in the form, the types tree will be a true dropdown/popover over existing content. This will require "unstyling" some of the things you did to make it lie directly on the page (height: 100%, etc.)
- Take a prop for whether it's a popover or not ^^
- Some additional logic changes: you might want to make parent components unselectable, because it shouldn't make sense for the user to easily select a ton of types. They should select one in particular. Clarify with Siraj

## Useful links

- Dow Jones documentation
- Existing Falling Fruit implementation of selecting types in a new location

laurenho025:
## Status:
:rocket: Ready

## Description
Adds the form type selection component as a flattened list of types using [react-select](https://react-select.com/home). The menu list component is replaced with [react-window](https://github.com/bvaughn/react-window) to handle the large amounts of type options.

Fixes #59 

## Screenshots
image


src/components/filter/TreeSelect.js
```diff
@@ -103,13 +135,16 @@ const TreeSelectContainer = styled.div`
...
     <NonrenderTreeSelect
       data={data}
       onChange={onChange}
-      texts={{ inlineSearchPlaceholder: 'Search for a type...' }}
-      showDropdown="always"
+      texts={{
+        placeholder: 'Select a type...',
+        inlineSearchPlaceholder: 'Search for a type...',
+      }}
+      showDropdown={popover ? undefined : 'always'}
```
Qrtn, line 147: `popover ? 'default' : 'always'` might be more clear.

src/components/mobile/SettingsPage.js
```diff
@@ -1,11 +1,13 @@
 import { SearchAlt2 as Search } from '@styled-icons/boxicons-regular'
 import { Cog, Flag, Star } from '@styled-icons/boxicons-solid'
-import { useState } from 'react'
+import React, { useEffect, useState } from 'react'
```
Qrtn, line 3: You can skip importing `React` in general in this repo.

src/components/ui/Select.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
```
Qrtn, line 1: Same here.

src/components/ui/Select.js
```diff
@@ -0,0 +1,76 @@
...
+    itemSize={LIST_ITEM_HEIGHT}
+  >
+    {({ index, style }) => <div style={style}>{children[index]}</div>}
+  </List>
+)
+
+const SelectWrapper = ({ options, placeholder, onChange, isMulti }) => (
+  <StyledSelect
+    components={{ MenuList }}
+    className="select-container"
```
Qrtn, line 63: Curious, what is the purpose of this className?

src/components/ui/Select.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react'
+import Select, { createFilter } from 'react-select'
+import { FixedSizeList as List } from 'react-window'
```
Qrtn, line 3: It might be more clear actually to not rename this import.

src/components/ui/Select.js
```diff
@@ -0,0 +1,76 @@
...
+  <List
+    height={maxHeight}
+    itemCount={children.length}
+    itemSize={LIST_ITEM_HEIGHT}
+  >
+    {({ index, style }) => <div style={style}>{children[index]}</div>}
+  </List>
+)
+
+const SelectWrapper = ({ options, placeholder, onChange, isMulti }) => (
```
Qrtn, line 60: I would suggest you just take props as an object, and spread them directly into StyledSelect. This allows users to specify additional props as well.

src/components/ui/Select.js
```diff
@@ -0,0 +1,76 @@
...
+    placeholder={placeholder}
+    onChange={onChange}
+    closeMenuOnSelect={false}
+    blurInputOnSelect={false}
+    // Reduces typing lag
+    filterOption={createFilter({ ignoreAccents: false })}
+  />
+)
+
+export { SelectWrapper }
```
Qrtn, line 76: I would export `SelectWrapper as Select`, to match the filename.

src/components/ui/Select.js
```diff
@@ -0,0 +1,76 @@
...
+    className="select-container"
+    classNamePrefix="select"
+    options={options}
+    isMulti={isMulti}
+    placeholder={placeholder}
+    onChange={onChange}
+    closeMenuOnSelect={false}
+    blurInputOnSelect={false}
+    // Reduces typing lag
+    filterOption={createFilter({ ignoreAccents: false })}
```
Qrtn, line 72: You can spread in `{...props}` at the end, allowing the user to override any of yours as they wish.

Qrtn (approved):
Great work Lauren!!

Qrtn:
Actually Lauren -- would you be able to style the select to look like the one in the Figma?

image

src/components/ui/Select.js
```diff
@@ -0,0 +1,76 @@
...
+  <StyledSelect
+    components={{ MenuList }}
+    className="select-container"
+    classNamePrefix="select"
+    options={options}
+    isMulti={isMulti}
+    placeholder={placeholder}
+    onChange={onChange}
+    closeMenuOnSelect={false}
+    blurInputOnSelect={false}
```
Qrtn, line 70: Pass these in from the parent rather than here. Otherwise, we can't use this component for single selects.

src/components/ui/Select.js
laurenho025: It assigns this classname to the outermost container of the select component. I thought I was going to use it for styling purposes but ended up not using it, so going to remove it.

src/components/mobile/SettingsPage.js
```diff
@@ -16,10 +18,39 @@ import { Slider } from '../ui/Slider'
...
   const [currentStep, setCurrentStep] = useState(2)
+  const [typeOptions, setTypeOptions] = useState([])
+
+  useEffect(() => {
+    async function fetchTypes() {
+      const types = await getTypes()
+      const options = types.map((t) => ({
+        value: t.id,
+        label: t.name,
+      }))
```
code-arman, line 31: make this like type instead of t?

src/components/mobile/SettingsPage.js
```diff
@@ -16,10 +18,39 @@ import { Slider } from '../ui/Slider'
...
+      setTypeOptions(options)
+    }
+    fetchTypes()
+  }, [])
+
+  const handleTypeSelect = (types) => {
+    const typeIds = types.map((t) => t.value)
+    console.log('Selected type IDs: ', typeIds)
```
code-arman, line 39: Remove printing before merging?

src/components/ui/Select.js
```diff
@@ -0,0 +1,97 @@
+import Select, { createFilter } from 'react-select'
+import { FixedSizeList } from 'react-window'
+import styled from 'styled-components/macro'
+
+const LIST_ITEM_HEIGHT = 46
+
+const StyledSelect = styled(Select)`
+  font-size: 18px;
```
code-arman, line 8: Is it supposed to be this big? The component seems to have a bigger font size and height than the other ones like the input box. Is this intentional?

code-arman (changes requested):
Looks really nice. Left some comments. Also have questions

Are there supposed to be duplicate types? 
Would it be good for there to be a X button to clear all the types?

src/components/mobile/SettingsPage.js
Qrtn: We can keep this here -- this kitchen page sink will eventually be removed.

src/components/ui/Select.js
Qrtn: ^ good question.

laurenho025:
> Looks really nice. Left some comments. Also have questions
> 
> Are there supposed to be duplicate types?
> Would it be good for there to be a X button to clear all the types?

Good questions. 

It looks like there's duplicate types because only the common name is used as the label. The current ff website includes the scientific name in brackets, but it makes the tags too long so for now siraj said to just display common names.

The figma didn't include a clear all button. @SirajChokshi @Qrtn thots?

src/components/ui/Select.js
laurenho025: Not really intentional, was just following the figma styling. Should I match it to the input box?

src/components/mobile/SettingsPage.js
laurenho025: hm maybe

src/components/ui/Select.js
Qrtn: Yes -- components are bigger on mobile than on desktop, but we'll handle that later. Also, can you style the text that comes up when searching to be same as the input box, too? Thanks.

Qrtn: Wait actually @SirajChokshi can you clarify, is there a difference between the heights of the select and the input?

Qrtn (approved):
Great work as always Lauren! Just check in with Siraj on the heights thing if you haven't already.

vasuchal (approved):
LGTM!!

code-arman (approved):
Great work as always Lauren!

PR #75, opened 2021-04-09
🧚🏿‍♂️ Form modal

vasuchal:
## Description
This is a modal that shows up when users want to fill out the form. 

Fixes #<number>

## Screenshots
Screen Shot 2021-04-08 at 9 42 27 PM

### DIFF
--- a/src/components/mobile/SettingsPage.js
+++ b/src/components/mobile/SettingsPage.js
@@ -1,9 +1,11 @@
 import { SearchAlt2 as Search } from '@styled-icons/boxicons-regular'
 import { Cog, Flag, Star } from '@styled-icons/boxicons-solid'
+import { useState } from 'react'
 
 import Button from '../ui/Button'
 import Input from '../ui/Input'
 import ListEntry from '../ui/ListEntry'
+import Modal from '../ui/Modal'
 import {
   AccordionItem,
   AccordionPanel,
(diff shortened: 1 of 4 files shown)


src/components/mobile/SettingsPage.js
```diff
@@ -16,6 +17,10 @@ import { Slider } from '../ui/Slider'
 import { Tag, TagList } from '../ui/Tag'
 
 const SettingsPage = () => {
+  const [showDialog, setShowDialog] = useState(false)
+  const open = () => setShowDialog(true)
+  const close = () => setShowDialog(false)
+  const buttonRef = useRef()
```
Qrtn, line 23: Why is this ref necessary?

src/components/mobile/SettingsPage.js
```diff
@@ -16,6 +17,10 @@ import { Slider } from '../ui/Slider'
 import { Tag, TagList } from '../ui/Tag'
 
 const SettingsPage = () => {
+  const [showDialog, setShowDialog] = useState(false)
+  const open = () => setShowDialog(true)
+  const close = () => setShowDialog(false)
```
Qrtn, line 22: You can pass these inline.

src/components/ui/Modal.js
```diff
@@ -0,0 +1,13 @@
+import '@reach/dialog/styles.css'
+
+import { Dialog } from '@reach/dialog'
+import styled from 'styled-components/macro'
+
+const StyledDialog = styled(Dialog)`
+  border-radius: 23px;
+  width: 75%;
+  height: 70%;
```
Qrtn, line 9: Is the padding correct on the dialog?

Qrtn (changes requested):
Nice work Vasu!

src/components/mobile/SettingsPage.js
vasuchal: I implemented the example from Reach UI to get the buttons to open and close the modal and they had refs but it seems to work without so I'll remove em!

src/components/ui/Modal.js
vasuchal: I looked on the figma and I couldn't find a padding property so I am not too sure how to handle this

Qrtn: Messaged you about this! But will include here as well for other ppl:

- A lot of layout CSS generated by Figma isn't too reliable, except for font sizes, shadows, colors, etc. Figma tends to generate absolute top, left, bottom, right measurements, which doesn't reflect that you should use padding
- Your goal is always to try to make your component **visually** similar to the design. Figma's CSS is just a tool to help you achieve that. Right now, it looks like there's too much space between the border and the text. How can you fix that?

src/components/ui/Modal.js
```diff
@@ -0,0 +1,13 @@
+import '@reach/dialog/styles.css'
+
+import { Dialog } from '@reach/dialog'
+import styled from 'styled-components/macro'
+
+const Modal = styled(Dialog)`
+  border-radius: 23px;
+  padding: 23px 17px 21px 17px;
```
Qrtn, line 8: The padding should probably be equal on the top and bottom.

src/components/ui/Modal.js
```diff
@@ -5,7 +5,7 @@ import styled from 'styled-components/macro'
 
 const Modal = styled(Dialog)`
   border-radius: 23px;
-  padding: 23px 17px 21px 17px;
+  padding: 23px 17px 23px 17px;
```
Qrtn, line 8: You can write this as `23px 17px`!

src/components/mobile/SettingsPage.js
```diff
@@ -16,6 +17,7 @@ import { Slider } from '../ui/Slider'
 import { Tag, TagList } from '../ui/Tag'
 
 const SettingsPage = () => {
+  const [showDialog, setShowDialog] = useState(false)
   const [currentStep, setCurrentStep] = useState(2)
```
Archna-1, line 21: What does the 2 mean  😮

Archna-1 (approved):
LGTM! :)

Qrtn (approved):
LGTM!

src/components/mobile/SettingsPage.js
vasuchal: it's the current step!

PR #89, opened 2021-04-15
🖖🏾 Settings Page

Archna-1:
## Description
Create settings page.

Fixes #<number>

## Todos
[ ] Add headers
[ ] Add checkbox for labels 
[ ] Add on/off slider for labels
[ ] Add buttons for map view


Qrtn (approved):
LGTM Archna!! Great job 😍

src/components/ui/ToggleSwitch.js
```diff
@@ -0,0 +1,63 @@
...
+    transition: 0.2s;
+  }
+`
+const CheckBox = styled.input`
+  opacity: 0;
+  z-index: 1;
+  border-radius: 15px;
+  width: 42px;
+  height: 26px;
+  &:checked + ${CheckBoxLabel} {
```
SirajChokshi, line 41: If I understand correctly this is supposed to style an active toggle label, can we reduce the styles below to only the essential changes (e.g. we probably don't need to redefine width, transition, height, etc)

SirajChokshi (approved):
Looking good! A few comments about code duplication and some cleanups, but I'll approve it now :)

PR #95, opened 2021-04-22
🏗️ Add TabsContext

Qrtn:
## Description

Move tabs info like index and setter to context.

Fixes something for Arman to merge


Qrtn:
Won't need this in lieu of new routing/deep-linking design in #98

PR #111, opened 2021-05-01
🏌️‍♂️ Add Entry Drawer and Deep Routes

### LINKED ISSUE #98: Routing/deep-linking
- Map view lives in `/map`
- Map view needs to store center and zoom in the URL at all times
  - Could look something like: `@40.110511,-88.2443827,14.94z`
- Selecting an entry now appends `/entry/ID` to the end of the current route
  - On map: `/map/entry/:ID/center_and_zoom`
    - Entry now appears selected on the map
  - On list: `/list/entry/:ID`
- Searching:
  - On map: `/map/search/:QUERY/center_and_zoom`
  - On list: `/list/search/:QUERY/center_and_zoom`
  - Searching removes the current entry
- In entry view
  - On map:
    - Clicking address: route doesn't change. Switches to drawer
  - On list:
    - Clicking address: route changes to `/map/entry/:ID`. Switches to drawer
  - Clicking map tab goes to `/map`, list tab goes to `/list` without `/entry` stuff
- Settings in `/settings` route

SirajChokshi:
## Status:

:rocket: Ready

## Description

Adds the following:
- Deep routing
  - Swap shallow routes via tabs
  - Replace links with new deep routes
- Drawer
  - Add drawer overlay and component

Fixes #98 

## Screenshots

tbd


src/utils/useRoutedTabs.js
```diff
@@ -7,15 +7,28 @@ import { useHistory, useLocation } from 'react-router-dom'
...
   const history = useHistory()
+
+  // get breadcrumbs
+  const segments = pathname.split('/')
+
   const [tabIndex, setTabIndex] = useState(() => {
     // Set the initial tabIndex from the URL on page load
-    const matchedIndex = tabPaths.indexOf(pathname)
+    const matchedIndex = tabPaths.findIndex((tabName) =>
+      pathname.includes(tabName),
```
Qrtn, line 17: I'd actually do `pathname.startsWith(tabName)`.

src/utils/useRoutedTabs.js
```diff
@@ -7,15 +7,28 @@ import { useHistory, useLocation } from 'react-router-dom'
 const useRoutedTabs = (tabPaths, defaultTabIndex = 0) => {
   const { pathname } = useLocation()
   const history = useHistory()
+
+  // get breadcrumbs
+  const segments = pathname.split('/')
```
Qrtn, line 12: Maybe do this split inside the else clause of `handleTabChange`?

src/components/mobile/TopBarSwitch.js
```diff
@@ -8,12 +8,12 @@ import TopBar from '../ui/TopBar'
 const TopBarSwitch = () => (
   <Switch>
     <Route path="/settings"></Route>
-    <Route path="/entry/new">
+    <Route path={['/map/entry/new']}>
```
Qrtn, line 11: Should we keep these as `="/map/entry/new"`?

src/components/mobile/MobileLayout.js
```diff
@@ -39,21 +40,29 @@ const MobileLayout = () => {
...
+                </Drawer>
+              </Route>
+            </Switch>
+            {tabPanels}
+          </Route>
         </Switch>
       </TabPanels>
       <Switch>
-        <Route path="/entry/new/details"></Route>
+        <Route path={['/map/entry/new/details']}></Route>
```
Qrtn, line 63: See TopBarSwitch.

src/components/mobile/MobileLayout.js
```diff
@@ -39,21 +40,29 @@ const MobileLayout = () => {
...
         <Route>
-          <TabList>{tabList}</TabList>
+          <TabList style={{ zIndex: 999 }}>{tabList}</TabList>
```
Qrtn, line 65: I would add a TODO here to 1) use the z-index enum and 2) override this in `ui/PageTabs` with a selector.

src/components/mobile/MobileLayout.js
```diff
@@ -39,21 +40,29 @@ const MobileLayout = () => {
...
             <LocationForm />
           </Route>
-          <Route path={['/map', '/list', '/settings', '/entry/new']}>
-            {tabPanels}
-          </Route>
-          <Route path="/entry/:id">
+          <Route path="/list/entry/:id">
             <EntryTabs />
           </Route>
+          <Route path={['/map', '/list', '/settings', '/map/entry/new']}>
```
Qrtn, line 49: I believe you don't actually need `/map/entry/new/details` here if the path is not exact and since it's in a switch? Not totally sure though.

src/components/mobile/MobileLayout.js
```diff
@@ -39,21 +40,29 @@ const MobileLayout = () => {
...
+            <Switch>
+              <Route path={'/map/entry/new'} />
```
Qrtn, line 51: Is this empty route and switch necessary if the parent route is already in a switch?

src/components/entry/EntryTabs.js
```diff
@@ -32,24 +34,28 @@ export const TextContent = styled.article`
...
-        </TabPanel>
-      </TabPanels>
-    </EntryTab>
-  </Page>
-)
+const EntryTabs = () => {
+  const { pathname } = useLocation()
+
+  return (
+    <Page isDrawer={pathname.split('/')[1] === 'map'}>
```
Qrtn, line 41: I think it's slightly better to pass `isDrawer` in as a prop from a routing-aware component (MobileLayout).

src/components/entry/EntryOverview.js
```diff
@@ -83,6 +83,9 @@ const EntryOverview = ({ className }) => {
   const { t } = useTranslation()
 
   useEffect(() => {
+    // clear location data when id changes
+    setLocationData(undefined)
```
Qrtn, line 87: Should we use `null` here?

src/components/entry/EntryNav.js
```diff
@@ -9,8 +9,8 @@ const EntryNav = () => {
   const { state } = useLocation()
 
   const onBackButtonClick = () => {
-    // Default to going back to the map. This occurs when the user opens /entry/{typeId} directly
-    history.push(state?.fromPage ?? '/map')
+    // Default to going back to the list. This occurs when the user opens /entry/{typeId} directly
+    history.push(state?.fromPage ?? '/list')
```
Qrtn, line 13: Can you add a TODO here to create a utils file for routing? It seems like we're doing a lot of manual fromPage sets and I would prefer to eventually make a set of standardized helper functions to do this. Thanks.

src/utils/useRoutedTabs.js
Qrtn: Actually Siraj, consider using `matchPath` from React router to handle this. https://reactrouter.com/web/api/matchPath

`exact: false, strict: false` (which are already defaults)

src/utils/useRoutedTabs.js
```diff
@@ -7,15 +7,28 @@ import { useHistory, useLocation } from 'react-router-dom'
...
+    )
     return matchedIndex === -1 ? defaultTabIndex : matchedIndex
   })
 
   const handleTabChange = (tabIndex) => {
     setTabIndex(tabIndex)
-    history.push(tabPaths[tabIndex])
+
+    // on shallow match return to shallow root
+    if (segments[1] === tabPaths[tabIndex].slice(1)) {
```
Qrtn, line 26: Same with matchPath here, with exact set to true. The reason is because `/mapasdf` could possibly match with `/map`.

src/components/entry/EntryTabs.js
Qrtn: Also, I think you should use `matchPath` for the condition. See my comments in `useRoutedTabs`

Qrtn (changes requested):
Beautiful work Siraj. I left some suggestions mostly on routing.

src/components/mobile/TopBarSwitch.js
SirajChokshi: Yeah, I'll make the change -- I kept a lot of these as arrays when developing since they were easier to change for testing stuff out.

Qrtn (approved):
Great work as usual dude. 💯

PR #147, opened 2021-05-24
🗾 Add list hovering and sync list results on map

### LINKED ISSUE #141: Hovering over list entry in desktop shows that it's selected on the map
## Learning objectives

- Objective

## Todo

These points are a rough guideline. **Please** feel free to discuss with others on the team about the best way to design this component!

- Todo

## Useful links

- Useful link

Qrtn:
## Status:
:rocket: Ready

## Description

Fixes #141 

## Screenshots

image

### DIFF
--- a/src/components/list/EntryList.js
+++ b/src/components/list/EntryList.js
@@ -31,9 +31,9 @@ const EntryList = forwardRef(
       itemSize,
       itemCount,
       locations,
-      handleEntryClick,
-      handleEntryMouseEnter,
-      handleEntryMouseLeave,
+      onEntryClick,
+      onEntryMouseEnter,
+      onEntryMouseLeave,
       ...props
     },
     ref,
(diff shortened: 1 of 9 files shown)


src/components/list/EntryList.js
```diff
@@ -31,7 +31,9 @@ const EntryList = forwardRef(
       itemSize,
       itemCount,
       locations,
-      handleListEntryClick,
+      handleEntryClick,
+      handleEntryMouseEnter,
```
SirajChokshi, line 35: Maybe we can just call these `onClick`, `onMouseEnter`, and `onMouseLeave`. No reason to reinvent the prop interface imo.

SirajChokshi (approved):
Clever Girl 🦖 
tenor

src/components/list/EntryList.js
Qrtn: Oh crap, yeah they should be on not handle lol.

Qrtn: Do you think `Entry` is necessary? Since it's for a specific item in the list.

SirajChokshi: Let's keep `Entry`. That's a good distinction to make.

ezwelty:
Don't forget to remove all locations (including those added from the list) from the map when back in cluster mode:

image

Qrtn:
Oh shoot, thanks Ethan.

PR #162, opened 2021-06-02
🚏 Add track location button

### LINKED ISSUE #131: Add track current location button and remove search entry
## Description

- Remove "Current Location" search entry
- Make current location button to the left of input on desktop do the same thing

## Learning objectives

- Objective

## Todo

These points are a rough guideline. **Please** feel free to discuss with others on the team about the best way to design this component!

- Todo

## Useful links

- Useful link

### LINKED ISSUE #163: Dropdown menus trigger mobile keyboard
**Describe the bug**
Dropdown menus trigger the mobile keyboard rather than the device's native select interface.

**To Reproduce**
Click on any dropdown menu. The text cursor blinks inside and (if on a mobile device) the keyboard opens.

**Expected behavior**
A text cursor does not appear, a keyboard does not open, and the device's native select interface is triggered. 
I suspect the easiest way to resolve this is to use a real `<select>` and style it with css as needed.

**Screenshots**
Screenshot_20210604-153528

Qrtn:
## Description

Add the ability to constantly track location of device. Both desktop and mobile track location buttons now behave the same, tracking the device's location at the center of the map.

Webpage now only requests for access to current location once a user action has occurred (i.e. clicking on the track location button).

Behavior mimics Google Maps.

Fixes #131 

Also fixes #163 in the settings page. For now, we'll default to disabling `isSearchable` when necessary since styling the native select is a bit tricky.

## Todos

- Fine-tune stop tracking location formula

## Screenshots

image
image
image


src/components/entry/EntryOverview.js
```diff
@@ -137,7 +117,12 @@ const EntryOverview = ({ locationData, className }) => {
 
             <IconBesideText bold onClick={handleAddressClick} tabIndex={0}>
               <Map color={theme.secondaryText} size={20} />
-              <LocationText>{address}</LocationText>
+              <LocationText>
+                {locationData.address ??
```
SirajChokshi, line 121: This is out of scope, but I'd prefer deconstructing all the attributes we need from `locationData` upfront.

src/components/map/ConnectedGeolocation.js
```diff
@@ -0,0 +1,16 @@
+import { useEffect } from 'react'
+import { useDispatch } from 'react-redux'
+import { useGeolocation } from 'react-use'
+
+import { geolocationChange } from '../../redux/mapSlice'
+
+export const ConnectedGeolocation = () => {
```
SirajChokshi, line 7: Is there a naming convention/surface-level distinction for utility components (i.e. components that provide state-related services, but don't render on the page) like this?

src/components/map/Map.js
```diff
@@ -36,14 +35,17 @@ const Map = ({
...
+      // tilt: 0,
     })}
     layerTypes={layerTypes}
     center={view.center}
     zoom={view.zoom}
     onChange={onViewChange}
     resetBoundsOnResize
   >
-    {!geolocation.loading && (
+    {geolocation && !geolocation?.loading && (
```
SirajChokshi, line 48: Can we get rid of the operational chaining on the second condition? 

`geolocation` will never be null nor undefined since we short-circuit on the first condition.

SirajChokshi (approved):
Left a few comments, but everything is functional/worth merging.

src/components/map/ConnectedGeolocation.js
Qrtn: Currently there isn't. I think this is a pretty rare case. I had to make this a component so that the webpage can defer requesting access to location by conditionally rendering ConnectedGeolocation.

PR #167, opened 2021-06-07
😩 Increase density of displayed clusters

ezwelty:
Just because the map feels a little empty when clusterZoom = mapZoom.

- Display {zoom + 1} clusters on {zoom} map.
- Make cluster background slightly transparent for when they overlap.

If we wanted to match fallingfruit.org exactly, it would be clusterZoom = mapZoom > 3 ? mapZoom + 1 : mapZoom. I think we did this to reduce overlaps for mapZooms 0-3. But I think it looks kinda cool with more stuff on the map.

## Status:

:rocket: Ready

## Description

More clusters!

## Screenshots

image

### DIFF
--- a/src/components/map/Cluster.js
+++ b/src/components/map/Cluster.js
@@ -24,7 +24,7 @@ const ClusterContainer = styled(ResetButton)`
   align-items: center;
   border-radius: 100%;
   color: ${({ theme }) => theme.background};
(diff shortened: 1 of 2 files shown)


Qrtn (approved):
Looks great, thanks Ethan. I do remember thinking that the map looked a bit sparse compared to the current website.

PR #213, opened 2021-11-09
👩🏿‍💼 Metric imperial setting

abehara2:
## Description

Fixes #<number>

## Screenshots

Screen Shot 2021-11-08 at 6 57 20 PM

Screen Shot 2021-11-08 at 6 58 25 PM

Screen Shot 2021-11-08 at 6 57 53 PM

### DIFF
--- a/src/components/list/EntryList.js
+++ b/src/components/list/EntryList.js
@@ -12,18 +12,19 @@ import ListEntry from '../ui/ListEntry'
 import { TypeName } from '../ui/TypeName'
 import { ReactComponent as LeafIcon } from './leaf.svg'
 
-const METERS_TO_FEET = 0.3048
-const FEET_TO_MILES = 5280
+const METERS_IN_FOOT = 0.3048
+const FEET_IN_MILE = 5280
 const CONVERSION_THRESHOLD = 1000
-const METERS_TO_KILOMETERS = 1000
+const METERS_IN_KILOMETER = 1000
+const IMPERIAL = 'imperial'
 
 const convertDistance = (distance, setting) => {
-  if (setting === 'imperial') {
-    const feet = Math.round(distance / METERS_TO_FEET)
+  if (setting === IMPERIAL) {
+    const feet = Math.round(distance / METERS_IN_FOOT)
     if (feet < CONVERSION_THRESHOLD) {
       return `${parseFloat(feet.toPrecision(2))} feet`
     } else {
-      return `${parseFloat((feet / FEET_TO_MILES).toPrecision(2))} miles`
+      return `${parseFloat((feet / FEET_IN_MILE).toPrecision(2))} miles`
     }
   } else {
     const meters = Math.round(distance)
@@ -31,7 +32,7 @@ const convertDistance = (distance, setting) => {
       return `${parseFloat(meters.toPrecision(2))} meters`
     } else {
       return `${parseFloat(
-        (meters / METERS_TO_KILOMETERS).toPrecision(2),
+        (meters / METERS_IN_KILOMETER).toPrecision(2),
       )} kilometers`
     }
   }
--- a/src/components/list/EntryList.js
+++ b/src/components/list/EntryList.js
@@ -2,6 +2,7 @@ import { ChevronRight } from '@styled-icons/boxicons-solid'
 import { darken } from 'polished'
 import { forwardRef } from 'react'
 import Skeleton from 'react-loading-skeleton'
+import { useSelector } from 'react-redux'
 import { FixedSizeList } from 'react-window'
 import styled from 'styled-components/macro'
 
@@ -13,6 +14,7 @@ import { ReactComponent as LeafIcon } from './leaf.svg'
 
 // TODO: use settings for this
 const convertMetersToMiles = (meters) => (meters * 0.000621371192).toFixed(2)
+const convertMetersToKM = (meters) => (meters / 1000.0).toFixed(3)
 
 const StyledListEntry = styled(ListEntry)`
   cursor: pointer;
@@ -35,6 +37,7 @@ const EntryList = forwardRef(
     { locations, onEntryClick, onEntryMouseEnter, onEntryMouseLeave, ...props },
     ref,
   ) => {
+    const settings = useSelector((state) => state.settings)
     const Item = ({ index, style }) => {
       const location = locations[index]
 
@@ -47,7 +50,11 @@ const EntryList = forwardRef(
             leftIcons={<EntryIcon imageSrc={location.photo?.thumb} />}
             rightIcons={<ChevronRight size="16" color={theme.blue} />}
             primaryText={<TypeName typeId={location.type_ids[0]} />}
-            secondaryText={`${convertMetersToMiles(location.distance)} miles`}
+            secondaryText={`${
+              settings.distanceUnit === 'imperial'
+                ? `${convertMetersToMiles(location.distance)} miles`
+                : `${convertMetersToKM(location.distance)} kilometers`
+            }`}
             onClick={(e) => onEntryClick?.(location.id, e)}
             onMouseEnter={(e) => onEntryMouseEnter?.(location.id, e)}
             onMouseLeave={(e) => onEntryMouseLeave?.(location.id, e)}
--- a/src/components/list/EntryList.js
+++ b/src/components/list/EntryList.js
@@ -58,7 +58,7 @@ const EntryList = forwardRef(
     { locations, onEntryClick, onEntryMouseEnter, onEntryMouseLeave, ...props },
     ref,
   ) => {
-    const { distanceUnit } = useSelector((state) => state.settings)
+    const distanceUnit = useSelector((state) => state.settings.distanceUnit)
     const Item = ({ index, style }) => {
       const location = locations[index]
 
--- a/src/components/list/EntryList.js
+++ b/src/components/list/EntryList.js
@@ -46,6 +46,7 @@ const EntryList = forwardRef(
     ref,
   ) => {
     const settings = useSelector((state) => state.settings)
+    console.log(settings)
     const Item = ({ index, style }) => {
       const location = locations[index]
 
--- a/src/components/list/EntryList.js
+++ b/src/components/list/EntryList.js
@@ -12,9 +12,17 @@ import ListEntry from '../ui/ListEntry'
 import { TypeName } from '../ui/TypeName'
 import { ReactComponent as LeafIcon } from './leaf.svg'
 
-// TODO: use settings for this
-const convertMetersToMiles = (meters) => (meters * 0.000621371192).toFixed(2)
-const convertMetersToKM = (meters) => (meters / 1000.0).toFixed(3)
+const convertDistance = (distance, setting) => {
+  if (setting === 'imperial') {
+    return `${(distance * 0.000621371192).toFixed(2)} miles`
+  } else {
+    if (distance >= 1000) {
+      return `${(distance / 1000.0).toFixed(3)} kilometers`
+    } else {
+      return `${~~distance} meters`
+    }
+  }
+}
 
 const StyledListEntry = styled(ListEntry)`
   cursor: pointer;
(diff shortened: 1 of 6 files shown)


src/components/list/EntryList.js
```diff
@@ -35,6 +45,8 @@ const EntryList = forwardRef(
     { locations, onEntryClick, onEntryMouseEnter, onEntryMouseLeave, ...props },
     ref,
   ) => {
+    const settings = useSelector((state) => state.settings)
+    console.log(settings)
```
Qrtn, line 49: Here you want to select `distanceUnit` directly. That way StyledListEntry won't just re-render on any settings change, only for distanceUnit.

src/components/list/EntryList.js
```diff
@@ -47,7 +59,10 @@ const EntryList = forwardRef(
             leftIcons={<EntryIcon imageSrc={location.photo?.thumb} />}
             rightIcons={<ChevronRight size="16" color={theme.blue} />}
             primaryText={<TypeName typeId={location.type_ids[0]} />}
-            secondaryText={`${convertMetersToMiles(location.distance)} miles`}
+            secondaryText={`${convertDistance(
+              location.distance,
+              settings.distanceUnit,
+            )}`}
```
Qrtn, line 65: I don't think you need to wrap this in backticks and dollar sign.

src/components/settings/SettingsPage.js
```diff
@@ -22,6 +23,11 @@ const LANGUAGE_OPTIONS = [
   { value: 'it', label: 'Italiano' },
 ]
 
+const METRICS_OPTIONS = {
+  metric: 0,
+  imperial: 1,
+}
```
Qrtn, line 29: Do you mean DISTANCE_UNIT_OPTIONS?

Qrtn: Would an array here work better?

```
[
  { value: 'metric', label: 'Metric' },
  { value: 'imperial', label: 'Imperial' },
]
```

Then you can pass that into `options`.

src/components/settings/SettingsPage.js
```diff
@@ -91,6 +104,16 @@ const SettingsPage = ({ desktop }) => {
           label={<label htmlFor={field}>{label}</label>}
         />
       ))}
+      <LabeledRow
+        label={<label htmlFor="distanceUnit">{t('Units')}</label>}
+        right={
+          <ButtonToggle
+            options={['Metric', 'Imperial']}
+            toggle={updateUnitsSetting}
+            selectedIndex={METRICS_OPTIONS[settings.distanceUnit]}
```
Qrtn, line 113: It's generally best to name these the standard `onChange`, `value` as in controlled React components.

src/components/ui/ButtonToggle.js
```diff
@@ -0,0 +1,44 @@
...
+  const renderButtons = () =>
+    options.map((option, index) => (
+      <Button
+        selected={index === selectedIndex}
+        onClick={() => toggle(index)}
+        key={option}
+      >
+        {option}
+      </Button>
+    ))
```
Qrtn, line 40: I think you can put this directly in ButtonContainer, no need to add an extra render function.

src/components/settings/SettingsPage.js
Qrtn: ButtonToggle could then return value directly in onChange.

src/components/settings/SettingsPage.js
```diff
@@ -58,6 +64,13 @@ const SettingsPage = ({ desktop }) => {
   const [overrideDataLanguage, setOverrideDataLanguage] = useState(false)
   const { t, i18n } = useTranslation()
 
+  const updateUnitsSetting = (index) => {
+    for (const [key, value] of Object.entries(METRICS_OPTIONS)) {
+      if (index === value) {
+        dispatch(updateSettings({ ['distanceUnit']: key }))
+      }
+    }
```
Qrtn, line 72: So you could avoid having this for loop, since onChange would return value directly.

src/components/settings/SettingsPage.js
```diff
@@ -58,6 +64,13 @@ const SettingsPage = ({ desktop }) => {
   const [overrideDataLanguage, setOverrideDataLanguage] = useState(false)
   const { t, i18n } = useTranslation()
 
+  const updateUnitsSetting = (index) => {
+    for (const [key, value] of Object.entries(METRICS_OPTIONS)) {
+      if (index === value) {
+        dispatch(updateSettings({ ['distanceUnit']: key }))
```
Qrtn, line 70: You actually don't need to put distanceUnit in quotes here. That's called a computed property name: https://ui.dev/computed-property-names/

It's already a string so you can just use it as a literal.

src/components/list/EntryList.js
```diff
@@ -11,8 +12,17 @@ import ListEntry from '../ui/ListEntry'
...
-// TODO: use settings for this
-const convertMetersToMiles = (meters) => (meters * 0.000621371192).toFixed(2)
+const convertDistance = (distance, setting) => {
+  if (setting === 'imperial') {
+    return `${(distance * 0.000621371192).toFixed(2)} miles`
+  } else {
+    if (distance >= 1000) {
+      return `${(distance / 1000.0).toFixed(3)} kilometers`
+    } else {
+      return `${~~distance} meters`
```
Qrtn, line 22: What's the tilde here do?

Qrtn (changes requested):
Nice work Ashank! Just a few things regarding component design with ButtonToggle.

src/components/ui/ButtonToggle.js
```diff
@@ -0,0 +1,44 @@
+import styled from 'styled-components/macro'
+
+const Button = styled.button`
+  background-color: ${(props) =>
+    props.selected ? ({ theme }) => theme.orange : '#ffffff'};
```
Qrtn, line 5: Is the colon supposed to go outside of the braces?

src/components/ui/ButtonToggle.js
```diff
@@ -0,0 +1,44 @@
+import styled from 'styled-components/macro'
+
+const Button = styled.button`
```
Qrtn, line 3: If you style `ResetButton` (which is in UI) here I think you can avoid some of the CSS properties.

src/components/list/EntryList.js
```diff
@@ -11,8 +12,17 @@ import ListEntry from '../ui/ListEntry'
 import { TypeName } from '../ui/TypeName'
 import { ReactComponent as LeafIcon } from './leaf.svg'
 
-// TODO: use settings for this
-const convertMetersToMiles = (meters) => (meters * 0.000621371192).toFixed(2)
+const convertDistance = (distance, setting) => {
+  if (setting === 'imperial') {
+    return `${(distance * 0.000621371192).toFixed(2)} miles`
```
AlbyYuggle, line 17: Would making this floating point number a const be better? Just so people understand what this number is.

src/components/list/EntryList.js
AlbyYuggle, line 49: Remove console log

ezwelty:
@abehara2 Nice work! However, may I suggest different units? I think distance should either be in feet (imperial) or meters (metric), or in miles (imperial) or kilometers (metric). Meters and miles are very different lengths.

Here is what the current mobile app uses – feel free to reuse. Integer meters up to 999 m, then 1.## km, or integer feet up to 999 ft, then 1.## mi.

```coffeescript
    distance_string: (meters)->
      if window.FFApp.metric
        meters = Math.round(meters)
        if meters < 1000
          return parseFloat((meters).toPrecision(2)) + " m"
        else
          return parseFloat((meters / 1000).toPrecision(2)) + " km"
      else
        feet = Math.round(meters / 0.3048)
        if feet < 1000
          return parseFloat((feet).toPrecision(2)) + " ft"
        else
          return parseFloat((feet / 5280).toPrecision(2)) + " mi"
```

src/components/settings/SettingsPage.js
abehara2: I tried the literal, but was thrown an error so will keep it as is.

src/components/list/EntryList.js
abehara2: Rounds to a whole number

src/components/list/EntryList.js
```diff
@@ -11,8 +12,30 @@ import ListEntry from '../ui/ListEntry'
...
+const CONVERSION_THRESHOLD = 1000
+const METERS_TO_KILOMETERS = 1000
+
+const convertDistance = (distance, setting) => {
+  if (setting === 'imperial') {
+    const feet = Math.round(distance / METERS_TO_FEET)
+    if (feet < CONVERSION_THRESHOLD) {
+      return `${parseFloat(feet.toPrecision(2))} feet`
+    } else {
+      return `${parseFloat((feet / FEET_TO_MILES).toPrecision(2))} miles`
```
Qrtn, line 26: Ashank, do you need parseFloats in all of these strings?

src/components/list/EntryList.js
```diff
@@ -35,6 +58,7 @@ const EntryList = forwardRef(
     { locations, onEntryClick, onEntryMouseEnter, onEntryMouseLeave, ...props },
     ref,
   ) => {
+    const { distanceUnit } = useSelector((state) => state.settings)
```
Qrtn, line 61: You should be more specific with the selector.

`const distanceUnit = useSelector((state) => state.settings.distanceUnit)`

src/components/settings/SettingsPage.js
```diff
@@ -58,6 +64,10 @@ const SettingsPage = ({ desktop }) => {
   const [overrideDataLanguage, setOverrideDataLanguage] = useState(false)
   const { t, i18n } = useTranslation()
 
+  const updateUnitsSetting = (object) => {
+    dispatch(updateSettings({ ['distanceUnit']: object.value }))
```
Qrtn, line 68: This can just be `dispatch(updateSettings({ distanceUnit: object.value }))`

src/components/ui/ButtonToggle.js
```diff
@@ -0,0 +1,44 @@
...
+  &:last-child {
+    border-radius: 0 3px 3px 0;
+  }
+`
+
+const ButtonContainer = styled.div`
+  display: flex;
+  flex: 1;
+  align-items: stretch;
+  width: 100%;
```
Qrtn, line 27: Is width 100% true by default?

src/redux/viewChange.js
```diff
@@ -95,6 +94,8 @@ export const viewChangeAndFetch = (newView) => (dispatch, getState) => {
     dispatch(stopTrackingLocation())
   }
 
+  // TODO: retrive event from viewChange and change order accordingly?
+
```
Qrtn, line 98: Is this todo still applicable?

Qrtn (approved):
Ashank, just a few more comments on this. Great work overall.

src/components/list/EntryList.js
abehara2: Tried refactoring this out, into different steps, but looks like it needs to be the last thing called for each value :/

src/redux/viewChange.js
abehara2: This was when I was drafting stuff for the optimization feature... whoops. Will remove this.

src/components/list/EntryList.js
```diff
@@ -11,8 +12,30 @@ import ListEntry from '../ui/ListEntry'
...
 
-// TODO: use settings for this
-const convertMetersToMiles = (meters) => (meters * 0.000621371192).toFixed(2)
+const METERS_TO_FEET = 0.3048
+const FEET_TO_MILES = 5280
+const CONVERSION_THRESHOLD = 1000
+const METERS_TO_KILOMETERS = 1000
+
+const convertDistance = (distance, setting) => {
+  if (setting === 'imperial') {
```
mattwalo32, line 21: Make a constant

src/components/list/EntryList.js
mattwalo32, line 26: I'd name this `NUM_FEET_IN_MILE` or something of a similar effect. When named `FEET_TO_MILES` it sounds like you should multiply by the constant instead of divide by it

mattwalo32 (approved):
Looks alright I guess

PR #215, opened 2021-11-09
🍞 Add toasts

### LINKED ISSUE #154: Add react-toastify for in-app notifications
**Describe the the feature request**

Add toasts for

- Successful add location
- Successful add review

code-arman:
## Status:
:rocket: Ready

## Description
Display toast on successful and unsuccessful location additions 

Fixes #154 

## Todos
- If you decrease the width of your screen to make it change to mobile view, the toast does not go underneath the search bar anymore. It only goes underneath when i chose a mobile screen size like iphone X

## Screenshots
Screen Shot 2021-11-12 at 3 20 50 PM

Screen Shot 2021-11-12 at 3 23 22 PM


src/App.js
```diff
@@ -26,7 +35,6 @@ const App = () => (
           <Route exact path="/login">
             <LoginPage />
           </Route>
-
```
SirajChokshi, line 29: tiny nit, but lets remove this change to reduce the diff

src/components/form/LocationForm.js
```diff
@@ -204,9 +205,13 @@ export const LocationForm = ({ desktop }) => {
...
 
-    console.log('locationValues', locationValues)
-    const locationResp = await addLocation(locationValues)
-    console.log('locationResp', locationResp)
+    let locationResp
+    try {
+      locationResp = await addLocation(locationValues)
+      toast.success('Location submitted successfully!')
+    } catch {
+      toast.error('Location submission failed.')
```
SirajChokshi, line 213: Maybe we could return a redirect to the /map page or something here on error since we could cause an error on line 223 (copied below) where we access an attribute of `locationResp`

```js
const reviewResp = await addReview(locationResp.id, reviewValues)
```

src/components/ui/Toast.js
```diff
@@ -0,0 +1,22 @@
+import { ToastContainer } from 'react-toastify'
+import styled from 'styled-components/macro'
+
+const Toast = styled(ToastContainer).attrs({
+  // custom props
```
SirajChokshi, line 5: delete this comment

src/components/ui/Toast.js
```diff
@@ -0,0 +1,22 @@
...
+  toastClassName: 'toast',
+})`
+  .Toastify {
+    &__toast {
+      size: 16px;
```
SirajChokshi, line 10: should this be font-size?

SirajChokshi (changes requested):
Good work! Few comments :)

src/components/form/LocationForm.js
code-arman: > Maybe we could return a redirect to the /map page or something here on error since we could cause an error on line 223 (copied below) where we access an attribute of locationResp

Wouldn't it be a weird experience if you submit a form and it redirects on failure? Like you'd probably want it to stay on the same @SirajChokshi

Qrtn (approved):

src/components/form/LocationForm.js
```diff
@@ -204,9 +199,16 @@ export const LocationForm = ({ desktop }) => {
...
+      history.push({
```
SirajChokshi, line 208: Maybe we could remove this push since the form data shouldn't be cleared out.

SirajChokshi (approved):
Looks good! Left one comment to delete a routing push

PR #256, opened 2022-01-05
📱 Mobile About and Authentication UI

SirajChokshi:
## Status:

:rocket: Ready

## Description

- Adds about routes as entries at the bottom of the Settings Page
- Adds information on the current user and the ability to logout (or a call to action to login or signup) at the top of the Settings page on mobile


Qrtn (approved):
Nice!

PR #267, opened 2022-01-06
🌾 Share the Harvest

SirajChokshi:
## Status:

:rocket: Ready

## Description

- Move generated data from `./constants` to `./constants/data`
- Move data script from `./genPressData.js` to `./scripts/data.js`
- Change build/script name to `yarn genData` from `yarn genPressData`
- Rewrote data saving script to be generic and handle both In the Harvest and Press Data datasets
- Added UI/Page for at `/share` for Share The Harvest data

## Screenshots
Screen Shot 2022-01-06 at 7 04 52 PM


scripts/data.js
```diff
@@ -0,0 +1,74 @@
...
+  output,
+  substitutions = {},
+  formatters = [],
+) {
+  const { data } = await axios.get(url)
+  let processed = data.map(substituteBy(substitutions)).filter(isIncluded)
+
+  formatters.forEach((formatter) => {
+    processed = formatter(processed)
+  })
```
Qrtn, line 62: reduce here instead of forEach

Qrtn (approved):
Very clean script!

PR #271, opened 2022-01-08
🖐️ Connect sign up form

### LINKED ISSUE #249: Clean up user sign up form
The user sign up form (https://falling-fruit.vercel.app/signup) needs to be reorganized to match the API.

- [x] Remove the two checkboxes. Editing anonymously is no longer an option while logged in, and we need to have the option to email users if the situation is important enough to warrant it.
- [x] Make `name` optional and move to above `bio`
- [x] Make `password_confirmation` invalid if empty and `password` is not empty
- [x] Require `password` to be 6 characters or longer (to match current mobile app)
- [x] Add reCaptcha

Currently, the API will return an error if the email is already taken. This means we need to show a (translated) message that this happened. However, it may be better to always send the email, but vary the content of the email based on whether the email was already taken or not – in which case, if the form is valid, it should always succeed.

Qrtn:
## Status:
:rocket: Ready

## Description

* Adds user sign up form
* Uses standard reCAPTCHA

Fixes #249 

## Todos

* Foraging range fields
* Translated message for duplicate email

## Screenshots

image

### DIFF
--- a/src/components/auth/AccountPage.js
+++ b/src/components/auth/AccountPage.js
@@ -83,7 +83,7 @@ const AccountPage = () => {
           initialValues={userToForm(user)}
           validationSchema={Yup.object({
             name: Yup.string(),
-            email: Yup.string().email(),
+            email: Yup.string().email().required(),
             bio: Yup.string(),
             new_password: Yup.string().min(6),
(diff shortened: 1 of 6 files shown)


SirajChokshi (approved):
Looks good :)

PR #280, opened 2022-01-10
🎛 Filter Render Delay

SirajChokshi:
## Status:

:rocket: Ready
## Description

Delays filter rendering until rest of map page mounts. Prevents seconds of delay when clicking the map page from another.


Qrtn (approved):
💯

PR #286, opened 2022-02-14
🐛 Various Bug Fixes

SirajChokshi:
## Status:

:rocket: Ready

## Description

- Fix check for if an entry has photos
- Fix validation step for seasonality
- Fix hidden close button for mobile street view
- Removed '?' icon next to empty review summaries
- Fixed bug where navigating to an entry and back would cause the filter to never load until a map movement
- Fixed bug where filter height would not scale to device height and cut off sidebar elements


Qrtn (approved):
good stuff

PR #287, opened 2022-02-14
🐛 Miscellaneous bug fixes

Qrtn:
## Status:
:rocket: Ready

## Description

Fixes:

- After clicking edit on an entry, the edit button stays around but doesn't do anything on click (not sure if this is intentional)
- After clicking settings and edit, a screen that says "Editing Location" shows up (not sure what is supposed to show up)
- Logging out on the accoutn page isn't reflected in the UI
- Can't log out of mobile. I tried clicking the logout button, but nothing would happen.
- No way to make a property private
- Seasonality dropdown is super small/difficult to use (looks like it's under a white pane)
- Edit button should not show if not logged in
- Unable to save a location after adding types. I clicked a location near me and tried adding some more types. It could not save and I got an error message. I was logged in.
- "Municipal" checkbox is not working
- "Invasive Species Only" button appears to have no function
- Also the icon of the self-location button turns into an X after clicking self-centering without giving access to location, and clicking on the X again will produce the same screen-zooming
- When clicking centering on self-location and access is not provided, clicking on the centering will just zoom in on the current screen(which I'm not sure if it's intended)

## Todos

- Convert forms on mobile to scrolling

### DIFF
--- a/src/components/auth/AccountPage.js
+++ b/src/components/auth/AccountPage.js
@@ -8,6 +8,7 @@ import * as Yup from 'yup'
 import { checkAuth, logout } from '../../redux/authSlice'
 import { editUser, getUser } from '../../utils/api'
 import { getPathWithMapState } from '../../utils/getInitialUrl'
+import { useAppHistory } from '../../utils/useAppHistory'
 import { PageScrollWrapper, PageTemplate } from '../about/PageTemplate'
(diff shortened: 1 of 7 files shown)


SirajChokshi (approved):
looks good 🚀

PR #329, opened 2022-12-13
🔧 Fix handling of types without English common names

ezwelty:
The API currently requires that a type have at least either an English common name or a scientific name. However, a type with only the former broke the building of the type tree, resulting is an empty tree. This pull request fixes this issue.

### DIFF
--- a/src/utils/buildTypeSchema.js
+++ b/src/utils/buildTypeSchema.js
@@ -83,8 +83,7 @@ const sortTypes = (types) =>
     )
 
 const getNames = (type) => {
-  const commonName =
-    type.name ?? (type.common_names.en && type.common_names.en[0])
+  const commonName = type.name ?? type.common_names.en?.[0]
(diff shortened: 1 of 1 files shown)


src/utils/buildTypeSchema.js
```diff
@@ -83,7 +83,8 @@ const sortTypes = (types) =>
     )
 
 const getNames = (type) => {
-  const commonName = type.name ?? type.common_names.en[0]
+  const commonName =
+    type.name ?? (type.common_names.en && type.common_names.en[0])
```
SirajChokshi, line 87:
```suggestion
    type.name ?? type.common_names.en?.[0]
```

SirajChokshi (approved):

src/utils/buildTypeSchema.js
ezwelty: @SirajChokshi Ah, very nice, thank you!

PR #330, opened 2022-12-14
👷🏼‍♀️ Omit authorization header when not needed

ezwelty:
I noticed that the HTTP Authorization request header is included on every request to the API, regardless of whether it is needed by the API or not. Besides potentially having security downsides (exposing credentials more than necessary), the presence of the header triggers CORS preflight requests for GET requests that would not otherwise trigger one, which has a performance impact.

I've hacked a solution whereby the header is omitted for select GET paths. But I wonder if this couldn't be generated automatically from the OpenAPI specification, where this is specified.


ezwelty:
Paths of public get requests are hardcoded for now. In the future, these could be scraped from the OpenAPI specification by a script and written to a JSON file imported by the app.

PR #346, opened 2022-12-28
🗑 Remove unused components, dependencies, etc

ezwelty:
In an effort to keep an overview of what is happening in the source code, and not add more work when refactoring, I think it would be helpful to remove any unused parts. Do you agree? They can always be added back if needed.

Here is what is removed thus far:

- [x] src/components/ui/ToggleSwitch.js
- [x] src/components/ui/SquareButton.js
- [x] src/components/list/PagedList.js
- [x] src/components/entry/PhotoGrid.js
- [x] src/components/about/TestPage.js
- [x] src/components/ComponentDemos.js (and `/demo` path)
- [x] src/components/about/CaptionInput.js
- [x] `react-dropdown-tree-select`

Is there anything else that is no longer being used in the visible interface that I've missed?


Qrtn (approved):
Hey @ezwelty, I think this is a good move, go for it!

PR #403, opened 2024-06-15
Remove custom styling of unselected tab icons

ezwelty:
I'm not sure why, but the List icon was the only icon affected by a custom color styling when unselected. Remove this style renders all of the tab icons the same (darker) color, which seems like an improvement.


wbazant:
Well spotted, I checked the list icon with and without the change and it does seem like an improvement!

PR #477, opened 2024-10-05
Fix layout issues around images on the About page for narrow screens (#471)

### LINKED ISSUE #471: Bad flow around images on about page with narrow screen
The about page (https://beta.fallingfruit.org/about) does not render well on narrow screens. Photos that are placed alongside text squash the text into a very narrow strip. The proposal is that below a threshold screen width, these photos should be moved to above their text so that the text can use the full page width.

marceloams:
Closes #471. In narrow screens (mobile), the images are now displayed above the text for improved readability.

### DIFF
--- a/src/components/about/PageTemplate.js
+++ b/src/components/about/PageTemplate.js
@@ -71,11 +71,33 @@ const PageTemplateWrapper = styled.article`
   }
 
   .content {
-    display: inline-flex;
-    align-items: flex-start;
     margin-top: 10px;
+
     p {
       margin: 0;
+    }
+
+    @media ${({ theme }) => theme.device.desktop} {
+      display: inline-flex;
+      align-items: flex-start;
+    }
+
+    @media ${({ theme }) => theme.device.mobile} {
+      overflow: hidden;
+      margin-top: 20px;
+
+      img {
+        max-width: 100%;
+        height: 100%;
+        float: left;
+        clear: left;
+      }
+    }
+
+    .div-aux {
+      float: left;
+      width: 0px;
+      height: 170px;
     }
   }
 `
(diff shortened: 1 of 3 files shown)


wbazant:
Thanks for making the changes, the text is more readable now! Would you be up for trying flex layouts?


Screenshot from 2024-10-05 08-51-42
The pictures don't need to be up to 160px if they're not to the side of the text, this 2x2 grid could take the whole width. 

I think this would look even better in a flex layout too:  the staff member's picture to the right, their name and email to the left, centered vertically, and jointly taking the whole width.
Screenshot from 2024-10-05 08-52-15

marceloams:
Hey @wbazant,

I've implemented the changes you requested for the grid. I also set a max height to prevent it from becoming too large, as you can see below:

Captura de tela de 2024-10-06 13-32-11

For the staff members section, this is the current result:

Captura de tela de 2024-10-06 13-05-09

However, I'd like to propose an alternative approach, where the image remains on the left and the text on the right. When the text exceeds the height of the image, it wraps underneath, taking up the full width, as shown below:

Captura de tela de 2024-10-06 13-26-26

I added a justify alignment for the text in this layout, but the left alignment works well too. Let me know your preference!

wbazant:
@marceloams thanks for making the grid take up the full width, it does look better on the screenshot. For the second part, I think your solution works better than what I proposed - once the director's name /title/ email fits on the right, then the text underneath will fit too. Would you push those changes to the branch?

marceloams:
@wbazant I've completed the requested changes. Please take a look when you have a moment, and I look forward to your feedback!

wbazant:
Thanks, that renders itself well on various screen sizes! I have an amendment to suggest but opening a PR to your fork felt like overkill here, so I merged your commit and opened https://github.com/falling-fruit/falling-fruit-web/pull/491 .

Would you be up for doing some other work on the project? The next thing to do for the About page is probably #492 , or you're welcome to grab any other issue.

PR #483, opened 2024-10-05
Add OSM and Stamen Toner-lite map layers

### LINKED ISSUE #352: Add OpenStreetMap map layer
Many users prefer OpenStreetMap as a basemap because it is much more detailed than Google Maps in some regions. See https://github.com/falling-fruit/falling-fruit/blob/master/app/assets/javascripts/mapcommon.js#L134 for implementation details.

ezwelty:
Closes #352.

* Adds the map layers used on the current website but still missing from the beta site.
* Adds the map layers to settings. If one of these is selected, all other controls specific to a Google Maps layer are hidden.
* Prints a custom attribution over the default at the bottom right of the map

Marked as draft because the attribution feels like a hack, but I'm not sure there is a better approach?


wbazant:
That will be a good addition!

There could be a better way of displaying these icons, maybe with flex grid - so that they stay the same size (rather than three squished and two full size), and on mobile five of them would probably fit in one row.
Screenshot from 2024-10-08 12-19-09
Screenshot from 2024-10-08 12-18-43

The beta design gives more space for the labels than the live site, so the labels could be longer and more informative. I checked "OpenStreetMap" and it looks quite crowded and I've eventually learned what it means. "Map" could be a longer label, because all options are maps. Maybe Google Map / Terrain / Satellit / OSM Base / OSM Toner Lite or so?

Might be a separate issue, but I noticed the "Biking, Transit" has the same UI as map tiles, but different UX - they look like a multiple choice but are effectively checkboxes. Also, I'm not sure what Transit does - at least near my house, the map has various bus and train stops on, and the option doesn't affect them.

The attribution partially covers Google's shortcuts and Terms. It's fine but maybe move  / remove those? It could also go instead of Google's link to the area in Google Maps in the bottom left corner.

ezwelty:
@wbazant Thanks for the feedback.

- I reorganized the tile buttons into "Google" and "OpenStreetMap" sections, dropping sections "Base" and "Overlay"
- I allowed the tiles to wrap, but this also required increasing the default width of the sidebar to 340 px (even after reducing the margins of the desktop settings page a bit). Frankly I feel the tiles are a waste of space, but l'd like to leave a redesign to some future time.
- I decided to leave Google-only features semi-transparent and disabled when OSM layer is selected, rather than hidden.
- Frankly I'm fine with Google basemaps and overlays as-is, because it matches the maps.google.com interface. Google Transit displays additional transit information – namely transit lines; check it out in Zürich ;)
- Removing the Google logo in bottom-left and the Terms on bottom-right is against the Google Maps terms of use, and there is no option through their API to alter the default copyright text displayed with custom layers: "Image may be subject to copyright". So I overlaid the required attribution for the custom layers on that line while leaving the Terms link still visible. 🤷🏼

wbazant (approved):
Looks good!

ezwelty:
It was just too dirty of a hack to put the attribution on the map. The results on narrow screens was lousy and the positioning would break as soon as we localize Google Maps and the text beneath changes.

So I moved it to the settings. This feels more robust.

PR #485, opened 2024-10-06
remove "by Anonymous" in anonymous reviews

### LINKED ISSUE #474: Remove "by Anonymous" in anonymous reviews
Reviews without an `author` are currently displayed as "Reviewed on {date} by Anonymous". This should be removed, so that the string reads "Reviewed on {date}".

https://beta.fallingfruit.org/locations/1824032

See `src/components/entry/Review.js`.

joaodantas91:
Closes #474.

### DIFF
--- a/src/components/entry/Review.js
+++ b/src/components/entry/Review.js
@@ -146,14 +146,9 @@ const Review = ({
       {!editable && (
         <cite>
           Reviewed on {formatISOString(review.created_at)}
-          {review.author && (
-            <>
-              {' '}
-              by {review.author}
-              {review.observed_on && (
-                <> (visited {formatISOString(review.observed_on)})</>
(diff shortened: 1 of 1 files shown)


ezwelty:
@joaodantas91 Thanks for jumping in! However, I'll note that #474 had already been assigned to @IvySaskia. Let's try to avoid stepping on each other's toes...

Regarding this PR, `observed_on` may be non-null even if `author` is null, so "visited ..." should still be displayed in that case. Please revise.

joaodantas91:
@ezwelty you're right, I'm so sorry about that!

I made the change so that observed_on can be displayed even if there is no user in the review.

wbazant:
Thanks, works as the issue requested!

PR #486, opened 2024-10-07
fix unexpected behavior of back button in side panel settings

### LINKED ISSUE #484: Unexpected back behavior in settings
On desktop, it is possible to have the settings tab open and move the map around. If one does so, clicking "<- Back" goes through each of the map views navigated to while settings was open, before closing the settings. I expect it to just close the settings sidebar?

Perhaps related: Clicking on a location while settings is open, then clicking back on location does not go back to settings as one might expect, but to the default map view.

badaminatasha:
From desktop: if settings page is opened in side panel, and user clicks on different locations on the map, the back button on the settings page goes through the location history before finally closing the settings page. Instead, it should leave the map as-is and only close the settings page. This bug was because the settings page back button used goBack function in onClick, whereas other similar components were using goToMap, which has the desired effect. Updated back button to use goToMap.

Closes #484

### DIFF
--- a/src/components/desktop/SidePane.js
+++ b/src/components/desktop/SidePane.js
@@ -63,7 +63,7 @@ const SidePane = () => {
     event.stopPropagation()
     history.push('/map')
   }
-  const goBack = (event) => {
+  const _goBack = (event) => {
     event.stopPropagation()
     history.goBack()
   }
@@ -122,7 +122,7 @@ const SidePane = () => {
             </Route>
             <Route path="/settings">
               <StyledNavBack>
-                <BackButton onClick={goBack}>
+                <BackButton onClick={goToMap}>
                   <ArrowBack />
                   {t('back')}
                 </BackButton>


badaminatasha:
Addresses #484

badaminatasha:
Closes #484

wbazant:
@badaminatasha thanks for making the change! Always going back to map follows user intention a bit better than going back in history, but we can access the settings page from a few more places - from my memory, the user could also be viewing a location, adding a location, editing a location, or editing a review.

A more complete implementation would use the state in Redux to figure out where to go back to - for example if there's a location id go back to the page of that location. Would you be interested in having a go? We do something like this for mobile, where being in settings and clicking a "map" icon does a few different things depending on the state. See implementation at https://github.com/falling-fruit/falling-fruit-web/blob/2437ae427242bd09fb6d19ca2eeac6ec85aab4aa/src/components/mobile/Tabs.js#L71 .

wbazant (changes requested):
Allow going back to not just the map page

wbazant:
Closed in favour of #545

Issue #383, opened 2024-05-26
Integrate location type categories


wbazant:
On live site, the "Show invasive species only" functionality is out of sight until zoomed in, and there are magic links leading to a portal with a different map.

 The current solution in the beta is two checkboxes
![Screenshot from 2024-05-26 16-43-37](https://github.com/falling-fruit/falling-fruit-web/assets/16976512/d1ba4d26-e9d8-4669-bf6e-28b63d3d497f) partially ported from live site. They are more prominent and I think they confuse the user - they get in the way, and offer explanations in pop-ups.

The magic links could be part of the UI and thus discoverable.

Some potential goals for the UX include:
- the default choice for categories stays out of the way until the user is ready (the non-default options like "Include municipal tree inventories" and "show invasive species only" are slightly further away from the UX of a first-time user)
- user learns to strategically choose groups of markers (e.g. decide to enable tree inventories to get more hits)
- discovery of categories introduces communities and lifestyle ideas (e.g. learning about that /grafting category might lead to someone trying grafting themselves)
- user can inform themselves on what the categories are

There's a list of all categories at #246. The categories are partially overlapping and also nested, suggesting a tree select, but that would take up space.

I considered some resolution options here, it seems like quite a UI/UX challenge and I don't feel I solved it, but a partial attempt is a labelled dropdown.
![Screenshot from 2024-05-26 19-29-18](https://github.com/falling-fruit/falling-fruit-web/assets/16976512/2f60be17-2b8a-42c2-8d86-61a1a648f573)

made by editing HTML in the browser to:
```
<div class="sc-ehIJor jTMUbL">
  <label for="options" class="sc-gnnDb eLamlu"><select id="options" name="options" class="sc-fFeiMQ">
      <option value="all">All</option>
      <option value="community_annotations">Forager + Freegan</option>
      <option value="community_annotations">Municipal Tree Annotations</option>
      <option value="community_annotations">Honeybee Plants</option>
      <option value="community_annotations">Forager</option>
      <option value="community_annotations">Freegan</option>
      <option value="community_annotations">Grafting</option>
      <option value="community_annotations">Community</option>
    </select>
    <span class="subtext" style="cursor:help;position:relative;top:-0.5em;">[?]</span>
  </label>
</div>
```
A multiselect is another partial option:
[...]

PR #494, opened 2024-10-09
Enhancement/default types

### LINKED ISSUE #467: Select only forager and freegan types by default
To match user expectations and the behavior of the current interfaces, the new `categories` property of each location type should be used to initialise the type tree filter with just those types selected.

frankgomezdev:
Closes #467. After implementing the changes, I used the data response from the browser's network tab to verify that the default selected options include 'forager' or 'freegan' as categories using the find function. Unfortunately, I did not check all of them.

Let me know if you'd like further adjustments!

image

### DIFF
--- a/src/components/filter/Filter.js
+++ b/src/components/filter/Filter.js
@@ -82,7 +82,9 @@ const Filter = () => {
             onSelectAllClick={() =>
               dispatch(
                 selectionChanged(
-                  typesAccess.selectableTypes().map((type) => type.id),
+                  typesAccess
+                    .selectableTypesWithCategories('forager', 'freegan')
+                    .map((type) => type.id),
                 ),
               )
             }
--- a/src/components/filter/Filter.js
+++ b/src/components/filter/Filter.js
@@ -82,9 +82,7 @@ const Filter = () => {
             onSelectAllClick={() =>
               dispatch(
                 selectionChanged(
-                  typesAccess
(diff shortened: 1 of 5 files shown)


wbazant:
@frankgomezdev Thanks! I've not ran the code yet, but from looking at the implementation, there seems to be bug for a filter indicator: if we uncheck Apple (forager category) but select Oak (not forager or freegan) we'll have non-default selection which misses a filter indicator because it's of the same length. Could you compare as sets?

wbazant:
Thanks @ezwelty  for making the issue and @frankgomezdev for doing this work! I knew about us needing the categories for the beta release, had vague ideas, and wasn't doing anything with them.

 @frankgomezdev if you'd like to do more work on the categories epic, I think the next step here is implementing them with a simple UI. I wrote up a plan in https://github.com/falling-fruit/falling-fruit-web/issues/383#issuecomment-2404655012  and can try to help you on the way, and you're also welcome to grab any other issue.

frankgomezdev:
That sounds good! I just took a peek, and steps 1 and 2 seem pretty manageable, but I might need some help with 3 and 4.

wbazant:
@frankgomezdev All right, we can try that! If you get stuck on the way, you can open a draft pull request.

There's also a non-implementation related difficulty to this task, we're not super sure how the UI should work - see the discussion - especially what the categories should do to the type tree - are they like search filters, or like the Select all / Deselect all button - but we can figure them out on the way.

Issue #416, opened 2024-07-03
Various small bugs and design suggestions for mobile layout


ezwelty:
Collected from myself and others over the recent past and dropping here for posterity. Can always spin off certain items as separate issues / pull requests.

* General
  * [ ] Make text larger
  * [ ] Use sentence case everywhere
  * [ ] Replace all "entry" with "location". Use "location" for a marked location on the map, and position for spatial coordinates. e.g. "Zoom into a location to see Entry Data" -> "Zoom in to list locations"
  * [ ] Use HTML dropdowns (`<select>`) for native UI on mobile
  * [ ] Fall back to English if no common name found in selected language. Adjust layout when no common name.
* Pages
  * [ ] Bad flow around images on about page with narrow screen
  * [ ] Include the Falling Fruit logo somewhere
  * [ ] Indicate whether user is logged in
  * [ ] Find a better location for the About pages than in the Settings
  * [ ] Data tables difficult to use on narrow screens. Could rows be collapsed somehow to reduce width?
* Map
  * [ ] Add small +/- zoom controls (see mobile app)
  * [ ] Make Current location and Add location buttons smaller and move off of map (see mobile app)
* Search header
  * [ ] Either use a straight bottom border to avoid hiding the top part of the map and reduce its height, or make the padding transparent (see Google Maps app)
* Filter pane
  * [ ] The filter badge number is distracting and should probably be removed - why 99+?
  * [ ] Changing the filter does not update the list
  * [ ] Make the filter pane fullscreen (extend to footer)
  * [ ] Combine filter with settings and convert to overlay drawer?
* List
  * [ ] Make photo thumbnails larger
  * [ ] Make photo thumbnail placeholder less prominent
* Location / Review
  * [ ] Use a straight top border for the location drawer
  * [ ] Make Fruiting status dropdown not searchable
  * [ ] Move the review summary to the location overview (or remove)
  * [ ] Replace "Anonymous" with no name
  * [ ] Banner photos are often too large (fuzzy) and the wrong aspect ratio. Replace with a horizontal scrolling gallery of fixed height but variable width.
* Street View
  * [ ] Close button is in the wrong location (too low)
  * [ ] Use same marker as on map

itzkushagra:
Hi, am beginner in open source can you assign me an easy good first issue amongst these to work.

ezwelty:
@itzkushagra Welcome, thanks for your interest! How about #471 ? (Bad flow around images on about page with narrow screen)

[...]

PR #501, opened 2024-10-11
fix: Street View close button location (#498)

### LINKED ISSUE #498: Close button for Google Street View is in the wrong location
The close button is too low on mobile. See for example https://beta.fallingfruit.org/locations/3383/panorama/@40.0406000,-105.2631900,16z

It is correctly located in the desktop view:

GooglyBlox:
### Happy Hacktoberfest! 
This issue fixes the top padding issue of the Google Street View close button. Sort of just a magic number solution, but it looks correct now.

Closes #498

### DIFF
--- a/src/components/map/CloseStreetView.js
+++ b/src/components/map/CloseStreetView.js
@@ -26,7 +26,7 @@ const StreetViewUIWrapper = styled.div`
   position: absolute;
 
   @media ${({ theme }) => theme.device.mobile} {
-    top: 100px;
+    top: 40px;
   }
 `
 
--- a/src/components/map/CloseStreetView.js
+++ b/src/components/map/CloseStreetView.js
@@ -24,10 +24,6 @@ const StreetViewUIWrapper = styled.div`
   left: 20px;
   justify-content: space-between;
   position: absolute;
-
-  @media ${({ theme }) => theme.device.mobile} {
-    top: 40px;
-  }
 `
 
 const CloseStreetView = () => {
--- a/src/components/map/MapPage.js
+++ b/src/components/map/MapPage.js
@@ -287,7 +287,14 @@ const MapPage = ({ isDesktop }) => {
       style={
         isDesktop
           ? { width: '100%', height: '100%', position: 'relative' }
-          : { width: '100%', position: 'fixed', bottom: '50px', top: '63px' }
+          : {
+              width: '100%',
+              position: 'absolute',
+              top: '48px',
+              bottom: '50px',
+              left: 0,
+              right: 0,
+            }
       }
     >
       {(mapIsLoading || locationIsLoading) && <BottomLeftLoadingIndicator />}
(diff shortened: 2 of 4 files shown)


ezwelty:
@GooglyBlox Thanks! Happy Hacktoberfest to you.

As you say, this is a magic-number solution. Upon reviewing your change, I realize that the custom styling on mobile is only needed because the search header is extending out onto the map:





This shouldn't be happening, because it could lead to confusing edge cases (e.g. the list shows locations but none are visible on the map).

Would you be willing to go deeper on this? https://github.com/falling-fruit/falling-fruit-web/issues/416 suggests using a straight bottom border for the search header. The map / street view container would then fill the space below the search header, removing the need for custom styling of the close button.

GooglyBlox:
I believe this commit should resolve the issue! Still uses some magic-number-esque solutions for handling some scaling, but I modified it to use a straight bottom border for the top bar, and then had it properly fill in the map underneath the top bar (which should prevent map items from being blocked). 

If there are any questions/comments/concerns you have, I'd have no problem working them through with you!

wbazant:
I like the corners being straight and the map being exactly under the top bar - no locations on the edge which are included in the list but not visible on the map, etc.

The change seems to introduce a subtle issue. When switching between Settings and Map I can see a top bar jiggle on my desktop setup for mobile (Firefox + Ubuntu + narrow window ). I can't reproduce on my 'actual mobile' setup ( Firefox + android + accessed over house wi-fi). Doesn't happen without the change.
https://github.com/user-attachments/assets/80609a65-c9d1-474a-8112-fec2afcf342d

@GooglyBlox Can you reproduce or have an idea why this happens? The styling of that top bar must be a bit dodgy, but the requirements for it are:
- 80 px at rest so that it matches the height of top ribbon in location drawer, so the UI switches between them exactly
- don't cover the map when at rest
- fold out when there are search results or when the filter is open

Also, there's an effect where the bar gets slightly bigger when the search input gains focus, and I'm not sure if we want it or not - it's kind of nice that after the user interacts with the element, the UI reacts and invites more interaction, but the input also gets a border so that's hopefully inviting enough.

GooglyBlox:
Hm, that's really strange! I can't seem to reproduce the top bar jiggle issue on Chrome or Firefox (on Windows 10, with React Developer Tools). Also not noticing any discrepancies when accessing it via my iPhone on Safari (which would make sense given the Developer Tools issue). 

Not super sure why the bar gets slightly bigger on focus, but it is a little bit of a strange issue that I will fix ASAP!

wbazant:
I had a bathroom thought, the unstable top bar height is a separate issue. I suspect the empty suggestions list incorrectly taking more than zero space or changing the layout in some way when it comes into focus. If you want to fix it, or do other stuff on the project, please do, and meanwhile I'll put the changes so far into the main branch since they solve what they say they solve. Thanks again!

 I made a ticket #507 to trace the follow-up.

PR #503, opened 2024-10-11
fix: Inconsistent alignment in type tree (#461)

### LINKED ISSUE #461: Misaligned type tree entries in Safari
On my system (Mac OS 11.7.7), Safari (16.5) fails to correctly display the type tree. Chrome and Firefox do so correctly. The CSS styling should be revised slightly and tested on many browsers to confirm that it is displayed properly.

arvin-z:
Browsers automatically compute `font-size` for form elements, and this can differ between browsers. This caused the `1em` ToggleButton and `1em` ChevronSpace to have differing sizes in some browsers, causing alignment issues as seen in #461. This fix adds `font-size: inherit;` to ToggleButton and ChevronSpace, ensuring that they always have equal `font-size`, therefore fixing the alignment issue.

Closes #461


wbazant:
Thanks for the fix! I couldn't test on Safari - I have an 2015 mac with Safari but the version must be quite old, the map doesn't even load - but I notice your change also fixes checkbox sizes on Firefox.

beta:
Screenshot from 2024-10-12 10-16-22

 your change:
Screenshot from 2024-10-12 10-19-22

PR #508, opened 2024-10-14
fix: unstable height of top bar (#507)

### LINKED ISSUE #507: Unstable height of top bar
Follow-up to #501. Suspecting the empty list of suggestions coming in and out of focus.

GooglyBlox:
This PR fixes the issue of having an empty suggestions list pop-up (which is a result of #501) when you focus on the search bar. 

Closes #507

### DIFF
--- a/src/components/search/Search.js
+++ b/src/components/search/Search.js
@@ -213,33 +213,35 @@ const Search = (props) => {
           />
         )}
       </SearchBarContainer>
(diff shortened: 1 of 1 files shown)


wbazant:
That works - not for my 'only in dev tools and only happens personally to me' jiggle on load, but works for focus. Thanks for following up on this!

PR #523, opened 2024-10-18
Make Fruiting status dropdown not searchable

### LINKED ISSUE #499: Make Fruiting status dropdown not searchable
The fruiting status dropdown is searchable, despite only having three options. On mobile, this is especially annoying because it triggers the opening of the keyboard.

Proposal: Make the dropdown behave the same as the location access or seasonality start and end month dropdowns.

marceloams:
- Aligned behavior with the location access dropdown.
- Prevented the dropdown from triggering the keyboard on mobile.

Closes #499


wbazant:
I checked out the commit locally and the change seems to work as intended, thank you!

PR #527, opened 2024-10-18
#513 Fix bug. Review not shown after adding new location with review

### LINKED ISSUE #513: Adding a location with a review shows the location without the review
A user can optionally submit a first review when adding a location. However, only the location is then displayed, and a page refresh or reselecting the location is needed for the review to also appear. This seems like an easy thing to fix in theory because the response from `POST /locations` (from which the new location is rendered) also includes the newly-added review.

paskal98:
Updated the reducer to correctly handle the `reviews` field, ensuring it retains or updates submitted review instead of defaulting to empty array

Closes #513

### DIFF
--- a/src/redux/locationSlice.js
+++ b/src/redux/locationSlice.js
@@ -239,7 +239,7 @@ const locationSlice = createSlice({
          * New location added
          */
         state.location = action.payload
-        state.reviews = []
+        state.reviews = action.payload.reviews || []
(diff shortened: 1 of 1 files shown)


ezwelty:
@paskal98 Brilliant! Tested and works. Nice when the fix is so simple.

(I also deleted all of your test locations).

PR #541, opened 2024-10-25
Allow changing position when adding location

### LINKED ISSUE #519: Location position matches searched place rather than chosen position
Steps to reproduce (mobile only, according to my tests):

- Search for a place and select from dropdown
- Zoom in or out (panning the map, or clearing the search, avoids the bug)
- Start add location
- Move location to a new position
- Submit the location
- **Location position is not the selected position, but that of the searched place!**

### LINKED ISSUE #539: Position change not respected when adding location
I discovered a (new?) serious bug that causes locations to be added at the wrong position. Applies to mobile only. Steps to reproduce:

- Navigate to a location: https://beta.fallingfruit.org/map/@64.2769622,-119.1532010,14z
- Begin adding a location: https://beta.fallingfruit.org/locations/new/@64.2769622,-119.1532010,14z
- Move the map
- Confirm. Note that the location position is not the final selected one but the original one: 64.276962, -119.153201 (from ...new/@64.2769622,-119.1532010,14z)

wbazant:
Closes #539 
Closes #519

Thanks @ezwelty for catching this, it was a high severity bug!

### DIFF
--- a/src/redux/locationSlice.js
+++ b/src/redux/locationSlice.js
@@ -10,6 +10,7 @@ import {
   getLocationById,
 } from '../utils/api'
 import { fetchReviewData } from './reviewSlice'
+import { updateLastMapView } from './viewportSlice'
 
 export const fetchLocationData = createAsyncThunk(
   'location/fetchLocationData',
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant Thanks for the quick fix!

PR #559, opened 2024-11-01
Autofix: 'On the map' checkbox causes a needless location query

lattaai1:
I've identified the issue and made the necessary changes to prevent the 'On the map' checkbox from reloading the map. The problem was in the Filter component where the 'showOnlyOnMap' checkbox was triggering a filter change that caused the map to reload. I've modified the Filter component to handle the 'showOnlyOnMap' checkbox separately from other filters that require map reloading.

Here's a summary of the changes:

1. In the Filter component, I separated the 'showOnlyOnMap' checkbox from other filters.
2. I created a new action 'setShowOnlyOnMap' in the filterSlice to handle the 'showOnlyOnMap' checkbox state change without triggering a map reload.
3. I updated the Filter component to dispatch the new 'setShowOnlyOnMap' action when the checkbox is toggled.
4. I modified the filterSlice to handle the new action and update the state accordingly.

These changes ensure that toggling the 'On the map' checkbox doesn't trigger a map reload while still updating the filter state. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

### DIFF
--- a/src/components/filter/Filter.js
+++ b/src/components/filter/Filter.js
@@ -5,6 +5,7 @@ import { useDispatch, useSelector } from 'react-redux'
 import styled from 'styled-components/macro'
 
 import { filtersChanged, selectionChanged } from '../../redux/viewChange'
+import { setShowOnlyOnMap } from '../../redux/filterSlice'
 import Input from '../ui/Input'
 import { CheckboxFilters } from './CheckboxFilters'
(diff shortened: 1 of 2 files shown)


wbazant:
@lattaai1 Please go away.

Issue #533, opened 2024-10-21
Make gap size between Sign In and Sign Up buttons not depend on the language


wbazant:
A bit too close in French, and maybe slightly too far in English:

![Screenshot from 2024-10-21 10-56-22](https://github.com/user-attachments/assets/9fca349f-aac7-4dfe-aed0-8640c5fbecf2)
![Screenshot from 2024-10-21 10-57-34](https://github.com/user-attachments/assets/2058b79d-17ec-4431-831d-9e9390caabb1)

I like the gap size between the review and report buttons, can we have that regardless of the language setting?

ahmedhalac:
Hi @wbazant / @ezwelty. Is this still open? Can i fix it?

PR #581, opened 2024-11-09
feat(#533): Changed gap size between signin  and signup button

ahmedhalac:
Resolving #533


wbazant:
Looks good, and I like the solution of making the button elements a sublist and adding a min-width and a fixed gap. Thanks for solving the issue!

PR #594, opened 2024-11-14
feat(#495): Change outdated twitter icon

### LINKED ISSUE #495: Outdated Twitter Icon (Bird Logo) Still Displayed After Rebranding to "X"
### Outdated Twitter Icon (Bird Logo) Still Displayed After Rebranding to "X"

**Issue Description:**
On the website, the Twitter icon is still showing the old bird logo instead of the updated X logo. 

### Steps to Reproduce

**Steps to Reproduce:**

1. Navigate to the [beta.fallingfruit.org](https://beta.fallingfruit.org/)
3. Observe the Twitter icon.

### Screenshots

image

ahmedhalac:
Closes #495


ezwelty (dismissed):
@ahmedhalac Thanks! Since we are going through the trouble, I would suggest embracing the X name more since it seems to be here to stay: 

* X-Twitter.svg → X.svg
* XLogo → X (to match the other social media icon components, if this is a valid component name)
* Update `alt` and `title` text from "Twitter" to "X"

Visually the X logo is less bold than the other social media icons we are using, but this feels like a detail.

wbazant:
@ahmedhalac can we add `<path stroke="#9b9b9b" stroke-width="0.7"` to the .svg? https://stackoverflow.com/a/37687733 told me that this is how one can make an .svg appear thicker, and it seems to work. 0.7 seems more or less right after trying a few values, but if you have a way to do it, you could set the value so it matches exactly the thickness of the instagram icon.

ahmedhalac:
Looks like this solved bold issue.

Now it looks perfect:

Screenshot 2024-11-17 at 5 03 04 PM

PR #537, opened 2024-10-23
Issue 496/add recent activity page

paskal98:
Added activity page with infinity scroll list 

Closes #496

PR #598, opened 2024-11-15
fix(#595): Fixed navbar content overflow for small devices

### LINKED ISSUE #595: Layout of top ribbon for small desktop heights can overflow
Minor issue, but there are widths where the current layout doesn't quite work. This is what a screen with 770px width looks like - note the Twitter icon going off, and a triangle in the 'about' button:
Screenshot from 2024-11-14 18-53-17

Screenshot from 2024-11-14 18-55-05

Maybe some clever layout changes will fix it - for example there's lots of space around the icon? If impossible, we can increase minimal width for desktop screens, e.g. from 767 to 800 px, but we'd like to understand how many pixels we need and for what.

ahmedhalac:
Fixed overflow for devices in width from 770px to 820px
Also fixed alignment for "Map" and "About" since they weren't aligned properly. There was minor difference.

TESTING RESULTS:

Screenshot 2024-11-15 at 10 28 53 PM
Screenshot 2024-11-15 at 10 29 33 PM

Closes #595

### DIFF
--- a/src/components/desktop/Header.js
+++ b/src/components/desktop/Header.js
@@ -51,9 +51,10 @@ const StyledHeader = styled.header`
     ul {
       list-style: none;
       padding: 0;
-      margin: 0 0 0 0;
+      margin: 0;
       height: 100%;
       display: flex;
+      align-items: center;
 
       li {
         display: flex;
@@ -66,6 +67,10 @@ const StyledHeader = styled.header`
         position: relative;
         font-weight: bold;
         font-size: 1rem;
+
+        @media (min-width: 770px) and (max-width: 820px) {
+          min-width: unset;
+        }
 
         a,
         .content button {
--- a/src/components/desktop/Header.js
+++ b/src/components/desktop/Header.js
@@ -68,7 +68,7 @@ const StyledHeader = styled.header`
         font-weight: bold;
         font-size: 1rem;
 
-        @media (min-width: 770px) and (max-width: 820px) {
+        @media (min-width: 767px) and (max-width: 820px) {
           min-width: unset;
         }
 


src/components/desktop/Header.js
```diff
@@ -51,9 +51,10 @@ const StyledHeader = styled.header`
     ul {
       list-style: none;
       padding: 0;
-      margin: 0 0 0 0;
+      margin: 0;
       height: 100%;
       display: flex;
+      align-items: center;
```
wbazant, line 57: This property makes the background look like this when Map is clicked:
Screenshot from 2024-11-16 15-44-41

src/components/desktop/Header.js
```diff
@@ -67,6 +68,10 @@ const StyledHeader = styled.header`
         font-weight: bold;
         font-size: 1rem;
 
+        @media (min-width: 767px) and (max-width: 820px) {
```
wbazant, line 71: I think you've correctly identified the reason for the layout to go funky, it's that min-width:110px that you're unsetting! I'm not sure if the solution is the best we can do, it happens to work in French, but at the cost of making the click area for the Map link quite small - see here with added background colors:
Screenshot from 2024-11-16 15-50-53

Is there a way to have a flexible layout for these? Also, there'll be a third link there - the Activity tab - see #537 . If you're working on this part of the layout, you could actually add the button and the verbiage, let it not go anywhere for now, and I'll meanwhile help finish that PR.

wbazant:
Closing in favour of #661. Thanks @ahmedhalac for figuring out what caused the overflow, it was helpful for solving the issue!

Issue #607, opened 2024-11-19
Replace Twitter bird with X logo on /sharing page


ezwelty:
The link to our X page was updated (https://github.com/falling-fruit/falling-fruit-web/issues/495), but I just realized that the same needs to be done on the sharing page, where we still display the old Twitter logo when linking to other organizations:
https://beta.fallingfruit.org/sharing

@ahmedhalac Interested 😉  ?

ahmedhalac:
@ezwelty sure. Will take this one.

PR #610, opened 2024-11-22
feat(#607): Changed outdated Twitter icon on share page

ahmedhalac:
Related to #607


ahmedhalac:
Testing result:

Screenshot 2024-11-22 at 1 20 08 AM

ezwelty:
@ahmedhalac Thanks for taking this on! Upon review, I realized that there were some other related issues with the code base (e.g. missing instagram icons, misuse of entry `RESOURCES` which is meant for location type rendering) and decided to build on what you'd started to address those as well. I also think the table looks nicer if we simple reuse the same icons that we use in the top bar. Do you agree?

ahmedhalac:
> @ahmedhalac Thanks for taking this on! Upon review, I realized that there were some other related issues with the code base (e.g. missing instagram icons, misuse of entry `RESOURCES` which is meant for location type rendering) and decided to build on what you'd started to address those as well. I also think the table looks nicer if we simple reuse the same icons that we use in the top bar. Do you agree?
> 
> 
> 
> 

Okay, I will check and get back to you with testing results.

ahmedhalac:
Hi @ezwelty. It looks much more cleaner now. Social icons in the table are matching navbar icons which is far better.

ezwelty:
> Do you need any changes from me here?

@ahmedhalac Nope, just merged. Thanks again!

PR #626, opened 2024-12-02
Fix round border from location drawer photo #602

### LINKED ISSUE #602: Remove round borders from location drawer photo
Note how the round borders leave a strange gap with the top of the window when the drawer is fully opened:

### LINKED ISSUE #627: Allow opening gallery from a half-open photo
We currently don't, but I don't see why not. Related to #602

christiancordeiro:
Removing the rounded from  location drawer photo

Closes #602
Closes #627

### DIFF
--- a/src/components/entry/Carousel.js
+++ b/src/components/entry/Carousel.js
@@ -38,10 +38,6 @@ const Carousel = styled(ResponsiveCarousel)`
       }
     }
   }
-  @media ${({ theme }) => theme.device.mobile} {
-    ${({ isFullScreen }) =>
-      !isFullScreen && `border-radius: 0 0; pointer-events: none;`}
-  }
 `
 
 Carousel.defaultProps = {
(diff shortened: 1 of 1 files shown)


wbazant:
Thanks for picking this up! Looking at your change and the code, I notice that:

1. isFullScreen is always false (nothing ever sets it to true)

2. You're also removing rounded corners from the half-open drawer, see how they look in the beta site here:
Screenshot from 2024-12-03 13-34-10

I'd be happy with that - I don't think the corners around the photo look very good, I think they made sense in a version of the layout where lots of other corners were rounded, and then probably we've not removed them.

3. The pointer-events: none has an effect we probably don't want

When removed, it makes the half-open drawer photo clickable, which I just tried on my phone and it seems good - see #627 to track.

The code change for 2. and 3. is simple - just remove the entire media clause - but would you check how it behaves, see if you agree with the intention or if you have other ideas?

4. I've found a related bug (#628) but that can be a different PR.

christiancordeiro:
I tested and removed the entire media clause and compared it with the current site. I didn't see any difference except for the removal of the rounded borders. I don't think there is any reason to keep this entire media clause. Additionally, it made opening the half-open gallery easier.

If so, can I make a commit by removing the media clause?

wbazant:
Let's change it like that then!

PR #638, opened 2024-12-10
Bump actions/cache@v2 to v4

ezwelty:
As announced (https://github.com/actions/cache/discussions/1510), actions/cache@v2 expires on 1 February 2015.


wbazant:
Let's try it!

PR #639, opened 2024-12-10
Allow clicks to pass between top buttons in mobile location page

ezwelty:
Could it be as simple as this?


wbazant:
Thanks for the fix! I checked it on desktop and mobile (Chrome/Firefox on Android) and it works as intended.

PR #690, opened 2025-02-13
Revise type search in location form

### LINKED ISSUE #684: Revisit the ignoring of characters in type search
Some of the last minute changes https://github.com/falling-fruit/falling-fruit-web/pull/641#issuecomment-2627749723

> Good suggestion! I ignored [^\w\s], so dashes, apostrophes, etc. I started ignoring space, until I realised I want the space as a feature - 'elm ' shouldn't match 'elmleaf blackberry' - so we allow word ends in input

My goal was to have name sets like `date-plum`, `date plum`, `dateplum` to be equivalent in search since these variations are so common. I guess that isn't possible if space is included (which I also see the value of in other contexts).

The larger issue is that removing `[^\w\s]` breaks the search for any name containing a letter other than a-z. For example, switch to French and search for "chne". It will match "chêne". Meanwhile, searching "chene" does not match "chêne" (tokenized as "chne"). Removing `[^\w\s]` after converting to ascii would address these issues with Latin diacritics, but still destroy searching in non-Latin scripts (e.g. good luck searching in Greek).

One fix would be to use Unicode character class escapes (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape). So `s.replace(/[^\s\w]/g, '')` would become `s.replace(/[^\p{L}\p{N}\s]/gu, '')` (anything that isn't a letter, number, or whitespace). An alternative would be to deliberately select the characters that we want to ignore, individually or using classes like `\p{P}` (punctuation), `\p{S}` (symbols), ...

If whitespace is preserved, it would probably make sense to squeeze all whitespace (`\s+`) to space (` `) and ignore any leading and trailing whitespace.

ezwelty:
After some experimenting and looking at common names in various languages, I opted to ignore whitespace. There are just so many cases of common synonyms differing by only a space, dash, or lack thereof.

Closes #684.

### DIFF
--- a/src/utils/localizedTypes.ts
+++ b/src/utils/localizedTypes.ts
@@ -101,7 +101,7 @@ const toMenuEntry = (
   }
 
   const commonNameLabel =
-    parentId === PENDING_ID ? `${commonName} (Pending Review)` : commonName
+    parentId === PENDING_ID ? `${commonName} (Pending review)` : commonName
 
   return {
     value: id,
--- a/src/utils/tokenize.ts
+++ b/src/utils/tokenize.ts
@@ -119,13 +119,19 @@ const addTokenStart: Transform = (s) => `${TOKEN_START}${s}`
 
 const addTokenStartEnd: Transform = (s) => `${TOKEN_START}${s}${WORD_END}`
 
-const rightTrimWithWordEnd: Transform = (s) => s.replace(/\s+$/, WORD_END)
-
 const tokenize = pipe(convertLatinToAscii, removeIgnoredChars, (x) =>
   x.toLowerCase(),
 )
 
-export const tokenizeReference = (strings: string[]): string =>
-  strings.map(pipe(tokenize, addTokenStartEnd)).join('')
+export const tokenizeReference = (strings: string[]): string => {
+  const tokens = strings.map(pipe(tokenize, (x) => x.replace(/ $/g, '')))
+  // Add tokens with spaces to the list with spaces removed
+  tokens.forEach((x) => {
+    if (x.includes(' ')) {
+      tokens.push(x.replace(/ /g, ''))
+    }
+  })
+  return tokens.map(addTokenStartEnd).join('')
+}
 
-export const tokenizeQuery = pipe(tokenize, addTokenStart, rightTrimWithWordEnd)
+export const tokenizeQuery = pipe(tokenize, addTokenStart)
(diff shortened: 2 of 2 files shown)


wbazant:
Collapsing references that differ only by a space is fine but removing all spaces disables something I was trying to achieve with 
```
const WORD_END = ' '
const addTokenStartEnd: Transform = (s) => `${TOKEN_START}${s}${WORD_END}`
const rightTrimWithWordEnd: Transform = (s) => s.replace(/\s+$/, WORD_END)
```
namely letting the user enter a space and remove some of the results. For example 'pear ' starts to match 'pearl oyster' with this change.

Do we still want to keep that feature, and make it so that omitting a space still gets the user a result, but typing a space removes unwanted results? ? We can e.g. include the references twice - with spaces removed and without.

ezwelty:
@wbazant I think we're going to have to choose. I've tried what you suggested (i.e., including the references twice) and `pearl oyster` is shown for `pearloyster` and `pearl oyster` (what I want) and not `pear ` (what you want). But there is still a problem. For example, search `small `. All the `small-...`, `small...` types disappear, and only reappear once an additional matching character is added after the space. Or search `leather ` and `leather leaf...` stays but `leatherleaf` and `leatherback` disappear from the results. To avoid this, I'd have to carefully curate the common names so as to always include a maximally-space-delimited form. Ignoring whitespace seems easier. Maybe take a test drive and let me know what you think.

ezwelty:
@wbazant This issue touches on all merged compound names, which is common in English (e.g. smallflower, redcedar) and even more rampant in German where this is the norm (e.g. Erdbeerbaum). Less an issue in Portuguese because the hyphenated form is almost always preferred.

For a curatorial solution: I could replace all of these with a hyphenated version. This wouldn't be ideal for display because sometimes one version (space, hyphen, or merged) really is more "correct" (and/or "common") than the others (e.g. elderberry). So instead, best would be, whenever the merged form is preferred, to always add a hyphenated version to the synonyms (which can't really be done automatically?).

It isn't unreasonable to emphasise curation, because other common search issues (gleaned from years of pending types) cannot be solved with clever regex. For example regional names (black[- ]?thorn vs. sloe), names of the plant vs the plant (rose[- ]?hip vs rose, hazel[- ]?nut vs hazel, poire vs poirier in French). The new search functionality can solve much historic confusion by including more synonyms.

wbazant:
Let's give up on the word boundaries idea - it just came from me as I tried to figure out how this should work, but I have a bias as an implementer of the feature. Thanks for considering it!

I didn't think to have the 'pending review' queue manageable as a design goal but that is indeed a benefit of ignoring white space - you only need to include the 'best' version as a name/synonym, but someone typing a less usual variant still gets a match.

PR #692, opened 2025-02-13
Update IRS links

### LINKED ISSUE #675: Fix 'closing remarks' on /about page
Here is the text:

Falling Fruit is a 501(c)(3) (tax-exempt) public charity based in Boulder, Colorado. As a result, donations within the United States are tax deductible. You may review our [letter of exemption](http://localhost:3000/501c3.pdf) from the IRS and verify our standing with [IRS Publication 78](http://apps.irs.gov/app/eos/pub78Search.do?) or the [Colorado Secretary of State](http://www.sos.state.co.us/biz/BusinessEntityCriteriaExt.do?resetTransTyp=Y).

- the letter of exemption is not migrated over from the live site
- the IRS publication link is bad

ezwelty:
Closes #675.


wbazant:
I see two problems:
- the old values are not updated so I'll have to be careful to not lose the change when I run --migrate
- this change only updates two languages but I'm planning to make more of them available

Options for now could be:
- making the change across all languages in Phrase, syncing to yaml files, and running the translation script which will update them here
- waiting for #695, asking the AI to make the change in all languages, checking the ones we can speak to see if they're correct, and assume it's done well everywhere if yes

ezwelty:
@wbazant Phrase doesn't seem to have search and replace capability. Since this isn't very high priority, I would suggest we come back to this soon after we disentagle translations from the legacy website.

PR #693, opened 2025-02-13
Remove or replace characters in French locale

ezwelty:
* Replace non-standard whitespace with ( )
* Replace non-standard apostrophe with (')
* Drop trailing whitespace

@wbazant Is it too early to deviate from Phrase now or is it irrelevant to your planned transition workflow?


wbazant:
See https://github.com/falling-fruit/falling-fruit-web/pull/692#issuecomment-2656365022 - can we make this change in the old .yaml and sync it? If that's not good, I can also add a bit of Python that will make it dynamically - then we still have the old files as the main source but correct values in this locale.

ezwelty:
See https://github.com/falling-fruit/falling-fruit-web/pull/692#issuecomment-2656634065

PR #696, opened 2025-02-13
Move desktop language select to header

### LINKED ISSUE #683: Move the language dropdown to the top bar
The language dropdown is currently in settings, which would be hard to find for someone opening the site and needing to change the language.

Proposing to move the dropdown to the top bar, similar to how it currently appears on the live site.

See https://github.com/falling-fruit/falling-fruit-web/issues/656#issuecomment-2614335606 for previous discussion / context.

ezwelty:
* I placed the select at the right-end of the main menu, so that it is accessed via hamburger menu in tight layout.
* I have no idea why it was necessary to wrap `LanguageSelect` in a styled div rather than style `LanguageSelect` directly.
* The additional element in the header meant increasing the max width of the tight desktop layout.

Closes #683.

### DIFF
--- a/src/components/desktop/Header.js
+++ b/src/components/desktop/Header.js
@@ -7,6 +7,7 @@ import { matchPath } from 'react-router'
 import { Link, NavLink, useLocation, useRouteMatch } from 'react-router-dom'
 import styled from 'styled-components/macro'
 
+import { LanguageSelect } from '../../i18n'
 import { logout } from '../../redux/authSlice'
 import { pathWithCurrentView, withFromPage } from '../../utils/appUrl'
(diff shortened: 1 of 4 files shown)


wbazant:
Thanks for implementingthis!

> I have no idea why it was necessary to wrap LanguageSelect in a styled div rather than style LanguageSelect directly.
Not sure, it doesn't seem to do anything - I removed it.

Also, I propose changing all select buttons throughout the site:
- the color should be theme.secondaryText instead of black
- no padding
Works better for this one but I think also e.g. fruiting status, since they don't need to be so big and puffy.

Issue #744, opened 2025-04-07
Observed on date not required for 'Flowers' fruiting status


ezwelty:
A bug that must have slipped in when form validation was improved. To replicate:

* Start a new location review
* Change fruiting status to "Flowers": Observed on is unchanged
* Change fruiting status to "Unripe fruit" or "Ripe fruit": Observed on is red as expected

My guess is that because "Flowers" has value 0, it is being treated as null.

p.s. Maybe once the legacy interfaces are down, I can replace all zero-based enumerables with integers >= 1 (or short descriptive strings). Because these 0 != null bugs have a tendency to sneak in.

aniyagnik:
Hi! I'm new to contributing to this project and would love to work on this issue. Please let me know if it's okay to proceed. Thanks!

ezwelty:
Welcome and yes, please have a go!

ezwelty:
Closed by #745.

PR #745, opened 2025-04-08
Fix observe state when selection 'Flowers'

Shrief88:
Update condition to handle undefined values instead of checking for null

solves [744](https://github.com/falling-fruit/falling-fruit-web/issues/744)


ezwelty:
@Shrief88 Thanks, that does the trick!

PR #754, opened 2025-04-11
Pass only the first detected language to i18next

### LINKED ISSUE #751: Inaccessible Chinese translations
Due to the by-default case-sensitivity of the `i18next` library, the Chinese translations are not accessible after switching to lowercase language codes.

### LINKED ISSUE #753: Wrong fallback behavior for Chinese locales
Although `zh-Hant` correctly resolves to our `zh-hant` translation, locales like `zh-Hant-{region}` or `zh-{region}` (for regions using traditional Chinese) resolve to `zh-hans`.

ezwelty:
Closes #753 .

Basically, I have to prevent an array of languages from being passed to `i18next.changeLanguage` for the fallbacks to work correctly. For testing, change `detection.order: ['navigator']`.

The behavior of i18next is so perplexing to me here that I made a bug report:
https://github.com/i18next/i18next/issues/2299


src/i18n.js
```diff
@@ -41,7 +41,9 @@ const LanguageSelect = () => {
   return (
     <Select
       options={LANGUAGE_OPTIONS}
-      value={LANGUAGE_OPTIONS.find((option) => option.value === i18n.language)}
+      value={LANGUAGE_OPTIONS.find(
```
wbazant, line 44: I didn't know i18n.languages is populated, does i18n.language ever have something different from i18n.languages[0]?

src/i18n.js
```diff
@@ -76,7 +78,15 @@ i18n
...
       lowerCaseLng: true,
-      fallbackLng: 'en',
+      fallbackLng: {
+        'zh-cn': ['zh-hans', 'en'],
+        'zh-my': ['zh-hans', 'en'],
+        'zh-sg': ['zh-hans', 'en'],
+        'zh-hk': ['zh-hant', 'en'],
+        'zh-mo': ['zh-hant', 'en'],
+        'zh-tw': ['zh-hant', 'en'],
+        default: ['en'],
```
wbazant, line 88: zh-hans and zh-hant are provided, not sure why you're providing the second fallback here?

src/i18n.js
```diff
@@ -87,5 +97,15 @@ i18n
     () => setDocumentDir(i18n.language),
   )
 
+i18n.on('initialized', () => {
+  // HACK: Pass single language to changeLanguage to avoid wrong fallback behavior
+  // 'zh-TW' -> ['zh-hant', 'en'] instead of ['zh-TW', ...] -> ['zh-hans', 'en']
+  let language = i18n.services.languageDetector.detect()
```
wbazant, line 103: Does this still happen if the fallback is provided as a string?

src/i18n.js
ezwelty: `i18n.language` is the current language (e.g. `zh-tw`), not the resolved language from the list of supported languages (e.g. `zh-hant`). i18n.languages are the selected supported languages (in fallback order) (e.g. [`zh-hant`, `en`].

src/i18n.js
ezwelty: Because if a custom fallback is provided, the default is not applied. So without this, you get `fr-fr` → `[fr, en]` but `zh-tw` → `[zh-hant]`.

src/i18n.js
ezwelty: Not relevant. See https://github.com/i18next/i18next/issues/2299#issuecomment-2796673234. It is arguably a bug in `i18next.services.languageUtils.getBestMatchFromCodes` called only if array is provided by detector, wherein it checks for `zh-TW`, then base language `zh` (ignoring script), then first derivative (`zh-Hans` in our `supportedLng` case).

ezwelty:
For a more robust fallback logic, but requires an additional dependency. It could be used with `convertDetectedLanguage` option (v7.1.0+ of i18next-browser-languagedetector):

```
const { LocaleMatcher } = require("@phensley/locale-matcher")

const localeMatcher = new LocaleMatcher(['en', 'zh-hans', 'zh-hant'])
var { locale, distance } = localeMatcher.match('zh-tw')
locale.id
// zh-hant
var { locale, distance } = localeMatcher.match('zh-hant-tw')
locale.id
// zh-hant
```

wbazant:
You seem to understand the situation better than me, we can go with whatever works! But it seems surprising that we need workarounds for what should be a common situation. If it's easier to name the locales zh-CN and zh-TW against the best practice doc but closer to what the browsers report that could be pragmatic too (the API can still refer to them properly).

ezwelty:
@wbazant Using `@phensley/locale-matcher` (see intermediate commit) is the robust way to go, but it adds a fair bit to the bundle size, so I went with a hacky i18next solution instead. https://github.com/i18next/i18next/releases/tag/v25.0.0 makes zh-Hant-TW → zh-Hant the default fallback behavior, but still is unable to do zh-TW → zh-Hant because `fallbackLng` is not invoked until after it strips down to the base language. Hence why I apply custom fallbacks to the value returned by the language detector instead. I couldn't get v25 to compile (probably needs some more modern compiler) so the custom fallbacks include both zh-Hant-TW (theoretically not needed with v25) and zh-TW versions.

wbazant:
Looks good! I've chosen Hong Kong Chinese plus some other languages in Chrome, cleared local storage, and had traditional characters show up:

Screenshot from 2025-04-16 14-08-02
Screenshot from 2025-04-16 14-08-07

```
wbazant@wbazant-19-01:~/dev/falling-fruit-web$ git diff
diff --git a/src/i18n.js b/src/i18n.js
index 8385d0ef..15c89594 100644
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -63,6 +63,7 @@ const LanguageSelect = () => {
   const { t, i18n } = useTranslation()
   const { googleMap } = useSelector((state) => state.map)
   const [hasToasted, setHasToasted] = useState(false)
+  console.log("LanguageSelect", i18n.language, i18n.languages)
   return (
     <Select
       options={LANGUAGE_OPTIONS}
@@ -98,6 +99,7 @@ i18n
         lookupLocalStorage: LANGUAGE_CACHE_KEY,
         caches: false,
         convertDetectedLanguage: (language) => {
+          console.log("convertDetectedLanguage", language)
           language = language.toLowerCase()
           return CUSTOM_FALLBACKS[language] || language
         },
```


We use i18n.language in lots of places in the code, and the select value appears to show up okay without doing i18n.languages[0] so I changed it back to i18n.language.

PR #801, opened 2025-05-17
Change how we load the Google Maps API

### LINKED ISSUE #800: Map does not load on iPad
The Google Maps element never loads on an iPad (iOS 14.8.1), tested on both Safari and Chrome. Only hint is this inscrutable error:

> Unhandled Promise Rejection: ReferenceError: Can't find variable: HTMLDialogElement

The problem is "fixed" by disabling the Places API. So it matches the issue identified here:
https://issuetracker.google.com/issues/375937548?pli=1

> The problem is reproducible with new map style for Google Maps Platform (v3.58) with Places API enabled. The root cause of the problem is that HTMLDialogElement is a relatively new HTML feature, and it is not supported in older browsers, including versions of Safari on iOS 15.2. The previous map style (v3.57) does not seem to have this problem and the Place Searches sample is displayed/works as expected.

Presumably updating iOS on this iPad would fix the problem. But this means that a lot of older iOS devices cannot use the website, which seems like a shame. So dropping to v=3.57 is a more inclusive fix.

ezwelty:
* Pin to version 3.57 (closes #800)
* Load asynchronously

### DIFF
--- a/src/components/map/MapPage.js
+++ b/src/components/map/MapPage.js
@@ -402,8 +402,9 @@ const MapPage = ({ isDesktop }) => {
         <GoogleMapWrapper
           onClick={handleNonspecificClick}
           bootstrapURLKeys={{
-            apiKey: process.env.REACT_APP_GOOGLE_MAPS_API_KEY,
-            version: 'beta',
+            // Maintain compatibility with older browsers
+            // https://issuetracker.google.com/issues/375937548?pli=1
+            version: '3.57',
             libraries: ['places'],
             language: i18n.language,
           }}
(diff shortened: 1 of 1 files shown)


wbazant (changes requested):
I'm surprised setting autoComplete to silence the warning is required. Can we set autoComplete to "on" for email fields, it's quite nice for the browser to autocomplete? I frequently have the email and password autocompleted for me on the login page.

I clicked around the map and don't see anything wrong with the async feature or the pinned version.

ezwelty:
@wbazant I'm fine with turning on autocomplete for the email on the sign in page, but I think you made a mistake and turned it on for the accountPage (where it is pre-filled and autocomplete would likely be a nuisance) when you probably meant the loginPage?

Issue #733, opened 2025-04-03
Improve mobile photo viewer: let modal fill entire screen and allow zooming into photos


wbazant:
I ran into a problem where I couldn't read from a label in the location https://beta.fallingfruit.org/locations/2111427/@55.8278571,-4.2507466,19z . 

Then I clicked into the picture, and it was better but I still couldn't read it:
![Image](https://github.com/user-attachments/assets/ede4308c-2f75-4e08-8ad5-81ad695edd4c)

I had what I needed only because I knew how to get to the image URL (https://ff-production.s3.us-west-2.amazonaws.com/photos/6be3cb8348bf0def63fe51ac024abd2d/original.jpg), that then lets you zoom in.

Proposed resolution: put an icon in the top right corner, next to X, that signifies full screen. When pressed, it could go to the image URL, but check that on swiping left or pressing 'back' in the browser, the user will return to where they want to be.

ezwelty:
@wbazant 👍🏼 This also bugs me. Along similar lines, I'd also suggest making the photo modal fill the entire screen (on mobile and desktop), as it seems a shame to have such big margins around what could be a bigger photo.

Daveside9:
Can I take this?

Issue #605, opened 2024-11-18
Improve banner photos on location page


wbazant:
We have this suggestion:

#416: Banner photos are often too large (fuzzy) and the wrong aspect ratio. Replace with a horizontal scrolling gallery of fixed height but variable width.

Also, photo orientation is sometimes not respected - a vertical photo might end up rotated ninety degrees.

wbazant:
Ethan says fuzziness is due to upsampling, 300px is less than typical phone width: https://github.com/falling-fruit/falling-fruit-web/pull/1085#issuecomment-4504788873 .

PR #820, opened 2025-05-31
Add side dots for carousel navigation in partially open drawer (#605, #733)

### LINKED ISSUE #741: Photo carousel: choose better signifiers for navigating between photos
Tiny issue for a small UX improvement! 

Compare a location with multiple photos - e.g. https://beta.fallingfruit.org/locations/2111424/@55.8276403,-4.2489323,18z - on desktop and mobile. On desktop or on a fully open drawer on mobile, there's a little row of circles in the bottom right corner, showing which part of a carousel you're on. But you can actually also click left and right in a partially open drawer, and navigate between photos, but the interface doesn't indicate it in any way.

We could pick different signifiers that will also work in a partially open drawer: maybe dots on the sides that are visible whenever you can swipe?

Clement-coder:
Added side dots to the photo carousel in EntryCarousel to indicate swipe navigation in a partially open drawer on mobile. This improves UX by providing visible navigation cues when the bottom dots are hidden.

- Added left and right side dots with click handlers for navigation.
- Hid bottom dots on mobile to avoid clutter in partially open drawer.
- Ensured side dots are only shown when there are multiple photos.

Closes #741


wbazant:
Thanks! I fixed your PR description - it seems you used a LLM to write the PR, and it claimed to close issues that your solution doesn't actually close.

1. Could you undo changes to files like package.json that this PR contains? Looks like you used npm to install it, and overwrote the setup based on yarn. Try follow the yarn based instructions to set up the project.

2. The new dots don't look good on desktop together with the old dots - can we have the old dots on desktop and new dots on mobile?

3. Can the photos be conceptually in a line, so we can go left and right, but not left when we're on the first photo and not right when we're on the last photo?

4. No need to make the dots themselves bigger, but can you make the click target a bit bigger, so the person using a phone with their thumb can easily aim at the dot and not the photo?

5. Can you use colors consistent with the rest of the app? see src/components/ui/GlobalStyle.js

6. After zooming into a picture, the dots go away and arrows appear instead. Can we offer a consistent way of navigating between the photos?

wbazant:
Thanks! No need to use lucid-react, if you want icons we have the boxicons library. What's your thinking behind changing dots to arrows?

wbazant:
Thanks! Feel free to take the time to do the changes. I've not seen how the latest version looks yet - I checked your first PR and now looked at the code - but before I do I think this is still to be addressed:

>  Can the photos be conceptually in a line, so we can go left and right, but not left when we're on the first photo and not right when we're on the last photo?

Can we implement this so that the 'go back' indicator is only visible on 2,3,4... photo, taking the user to the photo with one index less, and the 'go forward' indicator similarly? 'go back' is usually left and 'go forward' is right but we also have translations in Arabic and Hebrew where it's the opposite.

Also, see code review comments.

Clement-coder:
Fixed carousel navigation limits (no going past first/last image)

Renamed arrows for RTL support (BackArrow, ForwardArrow)

Removed unnecessary dependency (boxicons, etc)

Restored .env config

wbazant:
Thanks! Please also restore the yarn.lock file, we don't want those changes in this PR.

I checked out your code and ran it. The changes that only show the dot when there's something to show forward/backward seem to work in an LTR locale like English, thanks!

Can we improve the visual aspect of the solution? It looks like that on desktop, on mobile on half-open drawer, on mobile in fully open drawer, and on mobile in an open carousel:

Screenshot from 2025-06-03 09-12-22
Screenshot from 2025-06-03 09-13-14
Screenshot from 2025-06-03 09-13-40
Screenshot from 2025-06-03 09-16-28

I'm guessing it's just a result of LLM changes as opposed to a design? A designed solution would be nice - maybe it'll help if you draw in Paint or on a piece of paper what you'd want each of these views to look like first.

wbazant:
You can get your own for free I think - do the setup instructions mention that variable somewhere?

wbazant:
I looked at your code changes - changing PageTemplate.js wasn't right but otherwise I see you tried to improve how the dots look - and I ran your changes locally. Sorry to be closing the PR, but I appreciate the work you did - it helped me realise my suggested solution is probably not right here. Let's solve this issue differently - I've made #827, feel free to try it or take another issue.

PR #828, opened 2025-06-08
FEAT: added image that shows up while the app rendering

Thirukumaran-T:
Added a image tag  in the index.html which has a image to load up while rendering


Thirukumaran-T:
yes, i'll make the changes and update the PR

wbazant:
Thanks! Let's merge it in.

Issue #784, opened 2025-04-29
Remove unused code


ezwelty:
I notice that there are quite a few unused code elements lingering around. For example `Slider` and `ProgressBarStep`. A good approach would be to use a tool to find these programmatically:

* Upgrade project to Node.js v18.18.0 or higher
* Add `knip` (https://knip.dev/overview/getting-started)
* Set it up to lint the project

Thirukumaran-T:
Hi, I'd love to work on this  refactoring. Could you please assign it to me?

ezwelty:
@Thirukumaran-T Sure, go for it!

wbazant:
@Thirukumaran-T how are you doing with this issue? Were you able to clone the project and run it locally?

MonstersInc-sudo:
Hello,

I took a look at this issue and I am seeing breaking changes when upgrading to Node.js v18.18.0.

Instead of following the suggestion of adding it as a linter, I ran knip locally without installing and it returned a list of seemingly unused components.

Would you prefer that this issue be closed by keeping the node version the same and having items removed or upgrading project along with associated breaking changes to have the linter available going forward?

I would suggest we remove unused components(files) and delete unused dependencies from the package.json for the time being. 

Associated fork that does those recommendations. [Fork](https://github.com/MonstersInc-sudo/falling-fruit-web)

Attached is pdf of the output from knip.

[KnipOutput.pdf](https://github.com/user-attachments/files/21071560/KnipOutput.pdf)

Thanks,

Nic

wbazant:
Hi @MonstersInc-sudo , I looked at your fork, thanks for catching these - I looked through the list and it seems about right. Some of it was unfortunately me, looks like I renamed ConnectSearchParams and either didn't delete the old file or brought it back when merging. Feel free to submit a PR, and we can make a separate task to update node and add the linter.

wbazant:
Actually, don't remove ecosystem.config.js, Ethan added that and it looks like there's an intention behind the file being there.

PR #854, opened 2025-07-05
Partially fix issue #784

MonstersInc-sudo:
PR TO FIX ISSUE #784

- Removes all unused files as found by knip with the exception of ecosystem.config.js in which a comment was added.
- Remove dependencies that are unused. 
- Leave unused dev dependencies, and unadded dependencies used in code out of package.json.


wbazant:
Thanks for making the PR! Let me know if you want me to take over getting those changes in, but would you like to get the workflows green? The build is being slightly annoying, looks like  prettier wants to reformat your comment differently - we have it as a precommit hook, but if you e.g. added the change through the interface in github, it didn't run. Also, not sure why the branch doesn't cleanly rebase - probably rebasing locally will fix it.

PR #859, opened 2025-08-20
Capacitor

### LINKED ISSUE #704: Build mobile app from this web app (how?)
We have a mobile app for [Android](https://play.google.com/store/apps/details?id=uh.fallingfruit.app) and [iOS](https://apps.apple.com/us/app/falling-fruit/id380859409) which has been downloaded a combined 45 000 times. These apps have tens of thousands of users, but they are buggy, use a legacy Rails API, and are no longer maintained. This repo introduces a new mobile-friendly website which hopefully can be leveraged to replace the aging mobile apps. The question is _how_?

### Web

The legacy mobile app is a web app bundled as a mobile app using [Apache Cordova](https://cordova.apache.org). This tool is no longer maintained, but we could use its new descendent: [Ionic Capacitor](https://capacitorjs.com). My experience with Cordova suggests that this would be the easiest (if not the most performant) option, reusing nearly all the same code and requiring only a thin layer of platform-specific configuration.

My sense is that this is the only feasible option we have for quickly delivering mobile app users an updated experience.

### Native

Since this web app is written in React, it is tempting to suggest React Native, but these are very different beasts. They could share logic if as much as possible was moved out of the view components, but a migration is far from trivial. Something that could help is [React Native for Web](https://necolas.github.io/react-native-web/) or the superset dev platform [Expo](https://expo.dev/). But evidently it would have been easier to build a React Native app that also works for web than the other way around...

The third option is to start from scratch. But this would require a mobile dev team with long-term commitment to the project, otherwise we end up in the same situation that we've tried to escape: too much code and too little time.

eclaires:
Closes #704


wbazant:
I've read the set up docs and I appreciated they explained step by step what needs to be done and what happened when you built it, this is super helpful for #704 ! Happy to upgrade node on the whole project, but to merge the upgrade, the webapp needs to build as well. The github build going green could be the first target?

wbazant:
@eclaires Yes I think the github build is from scratch! the 'set up node' step says:

Acquiring 22.16.0 - x64 from https://github.com/actions/node-versions/releases/download/22.16.0-15177438473/node-22.16.0-linux-x64.tar.gz

Then the error is:
Error: [BABEL] /home/runner/work/falling-fruit-web/falling-fruit-web/src/index.js: Cannot find module '@babel/plugin-syntax-dynamic-import'

I've assumed you can read the logs, they're linked above under https://github.com/falling-fruit/falling-fruit-web/actions/runs/17111987250/job/48701366473?pr=859 and anyone should have visibility into it, but if not I can screenshot it for you. 


From what I understand, Babel is a transpiler, making source code javascript into 'javascript for browsers'. The error might comes from bumping react-scripts version in this PR too, which is meant to bundle Babel but something must be going wrong there.

Now that I think of what to do, this might be best path forward here:
0. start with non-upgraded Node and react-scripts
1. eject from react-scripts (https://create-react-app.dev/docs/available-scripts/#npm-run-eject ) which will replace our react-scripts dependency with Babel, webpack, etc. that it bundles
1.5. check our `yarn start` and `yarn build`
2. bump node versions one by one
3. arrive at a version which Capacitor is happy with

Issue #862, opened 2025-09-02
[URGENT] User email confirmation flow semi-broken


ezwelty:
We are receiving a spate of emails at info@fallingfruit.org from people who could not register. Investigating, I found the following.

## Email confirmation succeeds but errors

When the confirmation link is clicked (of the form):
https://beta.fallingfruit.org/users/confirmation?token={token}

The API is called twice in quick succession:
* POST https://beta.fallingfruit.org/api/0.3/user/confirmation?api_key=AKDJGHSD → 200 (toast: success!)
* POST https://beta.fallingfruit.org/api/0.3/user/confirmation?api_key=AKDJGHSD → 400 (toast: already confirmed...)

So it looks like it didn't work, because the error message is much larger and longer lasting than the success, and rather than end up on the sign in page with the email filled out, they end up at:
https://beta.fallingfruit.org/users/confirmation/new/@{latitude},{longitude},{zoom}z

## Missing form validation display

In the registration form, an empty password confirmation input does not appear red once a password is entered. So the Sign up button is disabled but it may not be clear why.

_Based on the number of recent actually-unconfirmed users, I suspect there is something else going on, but couldn't find what that might be._

wbazant:
We should fix the double message and the form, but also

> I suspect there is something else going on

Could it be a UX issue? I've once watched a friend follow this flow when doing beta testing and trying to add a location he's seen on holiday:
- search for falling fruit and then 'try the new website' as instructed
- end up at a sign in page
- I need to sign up. Go to the sign up page
- provide details, do the captcha, etc.
- see the map and a green success message
- Nice, I'm in! Find the location on the map
- Add the location without making an account

The two improvements that would make this not happen are my big 'Explore the map' button, so people who don't want to make an account don't feel like they need one, and #865. There could also be a technical issue causing this, like the emails taking a while or not being sent out that I don't know about, but that would be a separate thing!

PR #863, opened 2025-09-03
Fix signup flow

chrismaltais:
Closes: https://github.com/falling-fruit/falling-fruit-web/issues/862

The purpose of this PR is to fix the broken signup flow. Fixes by:
- Preventing duplicate API calls in ConfirmationPage due to React StrictMode
- Better handling of edge cases where users try to confirm an already-confirmed account
- Enhanced form validation for password confirmation that shows validation feedback immediately

I've implemented my best guess of the solution based on the issue description, however I don't have a great way of testing the signup flow.

Would love to learn how!

### DIFF
--- a/src/components/auth/ConfirmationPage.js
+++ b/src/components/auth/ConfirmationPage.js
@@ -1,4 +1,4 @@
-import { useEffect } from 'react'
+import { useEffect, useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import { toast } from 'react-toastify'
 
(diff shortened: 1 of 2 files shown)


wbazant:
I've now fixed the issue on #832, ended up using the invalidWhenUntouched solution from one of your changes with a slight tweak.

Sorry again for closing your PR - when I've read it, I've assumed it's entirely AI-generated without you having tested the code. When changes are partially right, I usually try suggest changes instead.

PR #869, opened 2025-09-16
Add missing photos

### LINKED ISSUE #864: Move press images over to the new repo
src/constants/data/press.json contains links to the old site, e.g. https://fallingfruit.org/press/earlytorise.jpg . Let's make sure these keep being served when we move over.

ezwelty:
Adds press and about (bio) photos. Closes #864.


wbazant:
Thanks, looks good! I checked they go into the `build` folder after `yarn build`, and that they display in dev.

PR #885, opened 2025-10-17
User's position in the list should be restored when going back to list from location page

### LINKED ISSUE #540: User's position in the list should be restored when going back to list from location page
Proposed in https://github.com/falling-fruit/falling-fruit-web/pull/525#issuecomment-2437115282. 

We have this working on activity pages - see `src/components/activity/RecentChangesPage.js` for how we `scrollIntoView`, and how we store the state in redux.

Implementation plan:
- when navigating to the location from the list, store an index in redux - for full correctness it should be the index of the element which is currently on top of the screen when it happens, but a clicked location's index is an alternative if the above is somehow impossible
- when rendering the list, use the stored index to navigate to an element

Marking as 'good first issue', this issue requires understanding how Redux and React work, but should be quite self contained otherwise!

marceloams:
### Changes Made

- Implemented scroll position restoration for locations list following RecentChangesPage pattern.
- Added Redux state management with lastViewedListPositionId in listSlice for position tracking.
- Created scroll restoration effect in ListPage using useEffect to restore position when returning from location pages.
- Added DisconnectLocationsPage component to clear scroll state when navigating outside /list and /locations context.
- Fixed layout issues by replacing scrollIntoView() with manual scroll calculation to prevent container shifting.

### Implementation Notes

Initial implementation used scrollIntoView() which caused the list container to shift up and go behind the fixed header. Solution was found on [Stack Overflow](https://stackoverflow.com/questions/11039885/scrollintoview-causing-the-whole-page-to-move) which recommended manual scroll calculation using offsetTop and scrollTop.

### Result

✅ Scroll position preserved when navigating list → location → back to list
✅ No layout disruption or container shifting behind header
✅ State properly cleared when navigating to unrelated pages
✅ Follows established project patterns from RecentChangesPage

Closes #540

### DIFF
--- a/src/components/mobile/MobileLayout.js
+++ b/src/components/mobile/MobileLayout.js
@@ -39,11 +39,11 @@ const MapContainer = styled.div`
 `
 
 const ListPageWrapper = styled.div`
-  height: 100%;
   overflow: scroll;
   margin-block-start: ${(props) =>
     props.isEmbed ? EMBED_HEADER_HEIGHT_PX : NAVIGATION_BAR_HEIGHT_PX}px;
-  padding-top: 4px;
+  inset-block-end: ${(props) => (props.isEmbed ? 0 : TABS_HEIGHT_PX)}px;
+  padding-top: ${(props) => (props.isEmbed ? 0 : 4)}px;
 `
 
 const shouldDisplayMapPage = (pathname) => {


marceloams:
Hey @wbazant! As discussed in the issue, could you please review this PR? Thanks!

src/components/connect/DisconnecLocationsPage.js
wbazant: Missing () after the action, did this work?

src/components/list/ListPage.js
```diff
@@ -30,6 +33,20 @@ const ListPage = () => {
     }
   }, [dispatch, locationsAvailable, locationsInvalid])
 
+  useEffect(() => {
+    if (lastViewedListPositionId) {
+      const locationElement = document.getElementById(lastViewedListPositionId)
+      if (locationElement) {
+        const scrollParent = locationElement.parentElement
+        if (scrollParent) {
```
wbazant: This will do if we can't make scrollIntoView work but we're losing some intelligent behaviour it has. Can you try modifying the styling of top component or putting a div around each location entry? I can have a go as well

src/components/connect/connectRoutes.js
```diff
@@ -273,6 +274,18 @@ const connectRoutes = [
     <DisconnectActivity />
   </Route>,
 
+  /*
+   * DisconnectLocationsPage
```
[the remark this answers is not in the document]
marceloams: Done, renamed it as proposed

src/components/list/ListPage.js
marceloams: I tried different approaches with `scrollIntoView()` but all of them caused the scroll container to shift slightly behind the header. The manual calculation with `offsetTop` and `scrollTop` is the only approach that worked. Feel free to give it a try if you have other ideas.

wbazant:
I think you've hit a bug we had since adding embed layout - going beyond the header when going back wasn't caused by scrolling into view, it was just always there, and there was also a missing offset at the bottom. Also, when describing my idea for DisconnectLastViewedListPosition, I forgot that there's also a /list-locations path we use between /list and /locations, so the component was unmounting and calling the clear() too early. Sorry about that! Have a look at my commits if you're interested in how I fixed those.

wbazant:
@marceloams I've made a follow-up issue, #888, about the scroll values.

By the way, feel free to submit PRs for anything you like in the project - you can comment under issues for help or solution ideas but there's no need for you to have them assigned. The solution descriptions can usually just say "closes #xyz" too.

PR #938, opened 2025-12-18
You are not signed in clickthrough

### LINKED ISSUE #642: Tell an anonymous user they are not signed in
Inspired by #632. When submitting locations/reviews anonymously, we ask the user to solve a captcha with the goal of discouraging vandalism. However, it's easy to forget to log in, and we don't explain the rule about only anonymous users having to fill out captchas, so there's potential for confusion.

This issue proposes adding text just under submit buttons in the forms (Location / Review / Add Type / Report):
"Submitting anonymously. CAPTCHA will be required".
or:
"You are not signed in. CAPTCHA will be required".

This will be especially nice after also solving #604 since then someone who's logged out but has an account can go back, log in, and come back to the form.

wbazant:
Closes #642 

Needs a small change on #899 to authStore / durable store (save 'do not ask again' as a Capacitor preference, delete from storage on logout).

### DIFF
--- a/src/components/form/NotSignedInClickthrough.js
+++ b/src/components/form/NotSignedInClickthrough.js
@@ -21,7 +21,6 @@ const StyledDialog = styled(Dialog)`
 `
 
 const Content = styled.div`
-  text-align: justify;
   margin-block-start: 1.5em;
   margin-block-end: 1.5em;
 `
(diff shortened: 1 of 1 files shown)


wbazant:
Needs an exception for the embed mode I forgot about - don't show the clickthrough in embed node

ezwelty:
@wbazant The quest for fewer accidentally-anonymous locations and more newsletter subscribers 😉 !

* [ ] [desktop] Not ideal for a "Review without signing it" modal to appear upon clicking a location or loading a location URL. Of course this should be limited to clicking the review button or latest the review submit button. Another reason to move the add review form on desktop to its own page/route once and for all?
* [ ] I could not get a modal to appear when reporting a location (bug?) or submitting a new type (intentional since within add/edit location form?).
* [ ] I'm getting a lot of uncaught timeout errors (at handleError (http://localhost:3000/static/js/bundle.js:137642:58), at http://localhost:3000/static/js/bundle.js:137665:7).
* [ ] Do not ask again → Don't ask again (matches the existing tone better?)
* [ ] I would suggest basing the modal title on the form title to make the intended action more clear.

{form title} anonymously?
Add location anonymously?
Edit location anonymously?

* [ ] The text feels too long. How about: "You are not signed in. Continue without an account, or first sign in (or sign up) to track and manage this observation." Information like "lets them appear with your name so others can see who did what" is better suited on the sign up and account page to explain public vs private profiles.

PR #947, opened 2026-01-20
Implemented pixel perfect restore for locations list

### LINKED ISSUE #888: Save and restore scroll position with pixels
Inspired by Marcelo's solution in #885. To restore positions when navigating back to list (and activity page) we currently remember the id of the clicked element, and use scrollIntoView to go back to it. This restores the view more or less correctly, but not quite to what it was.

We can actually do better, since we don't refetch the data between loads the layout of the components should just re-render to exactly the same heights unless the window resizes. So, we can remember three things:
- id of clicked element (typically location number)
- offset of the clicked element from the top of the container, element.offsetTop
- scroll position, parent.scrollTop

When restoring, check if the element exists, and if the offsetTop is the same as before, with accuracy of one pixel. If yes, set the scroll position exactly with scrollTop, otherwise fall back to scrollIntoView.

To test, try get back to the position of a clicked element in the middle of the screen. Then repeat, but make the window very narrow, which should make some heights different: observe the scrollIntoView behaviour.

Tiny enhancement!

jsamar17:
- Saved lastViewedOffsetTop and lastViewedScrollTop along with lastViewedPositionId.

- Renamed setLastViewedPositionId -> setLastViewedPositionState, clearLastViewedPositionId -> clearLastViewedPositionState.

- Implemented logic to save scroll position using element.offsetTop and element.parentElement.scrollTop.

- Retained fallback to scrollIntoView() in case of window resizing.

FIXES #888

### DIFF
--- a/src/components/list/Locations.js
+++ b/src/components/list/Locations.js
@@ -124,7 +124,7 @@ const Locations = ({
           lastViewedOffsetTop !== null &&
           locationElement.offsetTop === lastViewedOffsetTop
         ) {
-          parent.scrollTop = lastViewedScrollTop
+          locationElement.parentElement.scrollTop = lastViewedScrollTop
         } else {
           locationElement.scrollIntoView()
         }
@@ -162,12 +162,13 @@ const Locations = ({
           key={index}
           id={location.id}
           onClick={(e) => {
-            const locationPosition = e.currentTarget
-            onLocationClick({
+            const element = e.currentTarget
+            const locationPosition = {
               id: location.id,
-              offsetTop: locationPosition.offsetTop ?? 0,
-              scrollTop: locationPosition.parentElement?.scrollTop ?? 0,
-            })
+              offsetTop: element.offsetTop ?? 0,
+              scrollTop: element.parentElement?.scrollTop ?? 0,
+            }
+            onLocationClick(locationPosition)
           }}
         >
           <LeftIcon>


src/components/list/Locations.js
```diff
@@ -146,7 +163,17 @@ const Locations = ({
         <LocationItem
           key={index}
           id={location.id}
-          onClick={(e) => onLocationClick?.(location.id, e)}
+          onClick={(e) => {
+            if (e) {
+              return
+            }
```
wbazant: Can you explain this part? By reading the code, I guess (e) will be truthy, and it looks like onLocationClick will no longer work.

src/components/list/Locations.js
```diff
@@ -117,11 +119,26 @@ const Locations = ({
...
+        if (
+          lastViewedOffsetTop !== null &&
+          locationElement.offsetTop === lastViewedOffsetTop
+        ) {
+          const parent = locationElement.parentElement
+          if (parent) {
+            parent.scrollTop = lastViewedScrollTop
+          }
+        } else {
+          locationElement.scrollIntoView()
```
wbazant: Could you rewrite this conditional so that if locationElement.parentElement is missing, locationElement.scrollIntoView still happens? It will be clearer that it is a fallback

src/components/list/ListPage.js
```diff
@@ -59,12 +61,14 @@ const ListPage = () => {
...
+      lastViewedOffsetTop={lastViewedOffsetTop}
+      lastViewedScrollTop={lastViewedScrollTop}
       onClearLastViewedPosition={() =>
-        dispatch(clearLastViewedListPositionId())
+        dispatch(clearLastViewedListPositionState())
       }
-      onLocationClick={(locationId) => {
-        dispatch(setLastViewedListPositionId(locationId))
-        history.push(`/list-locations/${locationId}`)
+      onLocationClick={(scrollState) => {
```
wbazant: scrollState is not the right name for this variable, because it corresponds to a location. suggesting `locationPosition` instead

jsamar17:
I deleted a line in my earlier commit breaking the restore behavior. It's fixed now. Feel free to test it and tell me what you think

https://github.com/user-attachments/assets/4bfda566-7ce3-4ec1-a785-79a6d61e90f1

wbazant (approved):

Issue #807, opened 2025-05-24
Use HTML dropdowns (`<select>`) for native UI on mobile


wbazant:
react-select is very useful for our complex type dropdown and pretty on desktop, but not fully satisfying. On mobile, the native UI will offer a better experience to a user squinting at a screen or with cold thumbs. See https://github.com/falling-fruit/falling-fruit-web/pull/792#issuecomment-2901260613 for demo.

ezwelty:
@wbazant In the demo, the first row of photos is on desktop... So I'd go a step further to suggest that for simple dropdowns, a styled select element may be a general improvement, not just on mobile.

PR #948, opened 2026-01-22
Show labels more often

### LINKED ISSUE #940: Show labels more often
We currently show labels when:
- the user is adding a location
- 'on' setting is selected

I think we could rename the setting to 'always on' and intelligently decide whether we should show labels or not. Ideas for heuristics:
- zoom at least as large as geolocation zoom
- location density less than threshold (depends on the screen size)
- location number less than threshold (on desktop and mobile)
- show labels if there isn't a location just next to it on the screen

wbazant:
Closes #940

### DIFF
--- a/src/components/ui/LabeledRow.js
+++ b/src/components/ui/LabeledRow.js
@@ -18,6 +18,7 @@ const StyledLabeledInput = styled.div`
       font-weight: bold;
       color: ${({ theme }) => theme.secondaryText};
       cursor: pointer;
+      flex: 1;
     }
   }
 `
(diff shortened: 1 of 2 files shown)


ezwelty:
@wbazant I think the three options is a great solution.

* [ ] I would prefer keeping `labels=true` as always on, `labels=false` as always off, and only add a new value for zoomed in (e.g. `labels=zoomed_in` or `labels={x}z` where x is the z value at which labels appear) rather than overloading `true` with `always_on` and having to support two values for the same thing.
* [ ] I feel like this should be a select rather than radio buttons, such that it matches the existing Language and Units settings, and once we tackle #807 we will have a very nice native UI on mobile. Thoughts?

image

* [ ] It sounds strange for **Label _visibility_** to be "always on". I think it should be **Labels** or **Location labels**.

Fun fact, ArcGIS Online solves this with a slider to set the minimum (and in their case, also max) zoom level at which a layer is visible. But this goes in the opposite direction from our stated goal of seeking a default that works for everyone.

image

Issue #647, opened 2024-12-17
Add full access to Google Street View


ezwelty:
From a user:

> There's no longer the ability to start streetview at any point on the map by dropping the little guy onto it-- there's just the individual streetview links at each submitted point. Any hope of bringing that feature back?

If we want to match the behavior of the legacy website, from easier to harder:

- [ ] Enable the default Google Street View control (currently disabled because we override most default controls)
- [x] Draw all locations visible on the map also in Street View
- [ ] Show location details when clicking location marker in Street View (this used to work in the legacy website but seems to currently be broken)

wbazant:
An implementation comment, I found the Google Street view hard to work on (but adding locations and customising markers should be doable). And on to the wish list I could add toggling the street view when adding a location - it would be quite cool to be able to precisely add e.g. a roadside apple tree that's visible in the imagery.

I terms of planning and deciding how important this and other similar issues are, we'll get information by offering a link to the old website and keeping it up for a while. If many other users go back e.g. especially for general street view, then this issue can get a 'do before switching off the old website' milestone

ezwelty:
@wbazant Feedback from PortugueseFruitMan, with praise for the new website and a similar request for Street View:

> Today I tried the beta website:
> 1) I noticed that it is as fast as the legacy website in terms of scrolling, zooming in and out;
> 2) The beta website's new feature allowed me to add locations super fast;
> 3) I like the new 3D satellite view;
> 4) The only downside is that I couldn't find the option to see Google Street View on the beta website.

PR #955, opened 2026-01-30
Map with native markers

### LINKED ISSUE #142: Investigate drawing markers via native Google Maps API
(rewritten by @ezwelty)

The current website limits the map to 1000 locations per API call and 5000 markers total on the map -- which can be achieved by panning the map around in an area with a high location density to "paint in" additional markers (e.g. Denver, Colorado: https://www.fallingfruit.org/?z=13&y=39.72995&x=-104.98342). I don't notice a slowdown.

Yet calling `getLocations` with `limit: 1000` quickly leads to sluggish behavior, and `limit: 5000` crashes the page. To test (e.g. Denver, Colorado to test: http://localhost:3000/map/@39.7353078,-104.9682193,14z).

My suspicion is that adding markers via the Google Maps API directly using a PNG image is much more efficient than rendering each marker as an HTML element / React component.

https://developers.google.com/maps/documentation/javascript/markers

For example:

```js
new google.maps.Marker({
  icon: {
    url: "img/png/map-location-dot.png",
    size: new google.maps.Size(w, h),
    origin: new google.maps.Point(0, 0)
  },
  // by convention, icon center is at ~40%
  anchor: new google.maps.Point(w * 0.4, h * 0.4),
  position: new google.maps.LatLng(lat, lng),
  map: map,
  title: title,
  draggable: false
  zIndex: 0
})
```

wbazant:
Closes #142

### DIFF
--- a/src/components/entry/overview/TypesHeader.js
+++ b/src/components/entry/overview/TypesHeader.js
@@ -146,9 +146,7 @@ const TypesHeader = ({ types, openable }) => {
           // At least 1 URL
           return (
             <TypesAccordionItem key={type.id}>
-              <TypesAccordionButton openable={openable}>
-                {typeTitle}
-              </TypesAccordionButton>
+              <TypesAccordionButton>{typeTitle}</TypesAccordionButton>
               <TypesAccordionPanel>
                 <ResourceList urls={type.urls} />
               </TypesAccordionPanel>
--- a/src/components/map/SelectedLocation.js
+++ b/src/components/map/SelectedLocation.js
@@ -1,7 +1,7 @@
 import { BackgroundMapPin, MapPin } from './Pins'
 
-const SelectedLocation = ({ selected, editing, ...props }) => (
-  <div dir="ltr" {...props}>
+const SelectedLocation = ({ selected, editing }) => (
+  <div dir="ltr">
     {selected && !editing && <MapPin />}
     {editing && <BackgroundMapPin />}
   </div>


ezwelty:
@wbazant 

> I've swapped the .png for an .svg. Looks crisp and as far as I can tell, it still loads quickly.

I had expected that to be less performant, but it seems like the SVG is being drawn into the canvas like the PNG.

> I'm told it is a side effect of the marker being drawn onto the canvas. We can control it with a parameter optimized. How about optimized: true for over a hundred markers? I'm yet to fix what happens if you start far off and zoom in (a non-optimized marker stays non-optimized but could be tweaked).

No one ever complained about this on the old website, so maybe it's not too serious?

> I'll try bringing it back! Should we have a separate styling of labels for satellite view?

Rather one. The two styles I presented above seemed to work in a wide range of situations. Here was the code:

```js
div.style.position = 'absolute'
div.style.padding = '1px 4px'
div.style.fontSize = '0.875rem'
div.style.pointerEvents = 'none'
div.style.marginTop = '10px'
div.style.textAlign = 'center'

// version 1
div.style.fontWeight = 'bold'
div.style.color = theme.background
const color = theme.headerText
div.style.textShadow = `-1px -1px 0 ${color},
  1px -1px 0 ${color},
  -1px 1px 0 ${color},
  1px 1px 0 ${color}`

// version 2
// div.style.color = theme.headerText
// div.style.backgroundColor = 'rgba(255, 255, 255, 0.8)'
// div.style.borderRadius = '2px'
```

> I've not yet checked Street View but all the markers appearing are indeed good - the ultimate feature would be draggable markers, so you could put the marker exactly on the tree :) - but unless Google supports it, I have no idea how that would work.

Google supports it. Basically happens automatically if you have a draggable marker on the map and enter street view via the built-in control (which has been requested to bring back: #647). We have that on the old website. https://www.fallingfruit.org/locations/3383/edit, slide the street view person onto the map, and drag the marker. It's cool and can sometimes be useful, but if the vertical imagery is any good, it's often more accurate to use that instead.

p.s. I notice that clicking a marker to select it for the first time throws a bunch of console warnings:

Screenshot 2026-02-06 at 21 41 28

wbazant:
I've restored text shadow and removed the warnings. Let's merge it so I can build on top of it before this branch starts growing arms and legs. I will try the new styles for labels but seems easier in the next PR.

PR #960, opened 2026-02-07
Share links to the main site

### LINKED ISSUE #958: Coordinate base link for sharing in app
useShareUrl.js:
```
  const baseUrl = Capacitor.isNativePlatform()
    ? 'https://beta.fallingfruit.org'
    : window.location.origin
```

Let's think what to set it to before we update :). Should probably be fallingfruit.org if we're ready to release together.

### LINKED ISSUE #977: Config mobile app properties
For a smooth upgrade, the new app versions need to have the same identifier as the old one, a higher version number (0.3.0 to indicate upgrade to API version 0.3), and built with the same (minimum) operating system requirements so that no current user is kicked out by the upgrade. In the old Cordova `config.xml` format, these parameters would have looked like this:

```xml
<widget
  id="org.fallingfruit.app"
  version="0.3.0"
  android-versionCode="3000"
  android-packageName="uh.fallingfruit.app"
  ios-CFBundleIdentifier="com.fiddlemeragged.fallingfruit"
>
  <name>
    Falling Fruit
  </name>
  <description>
    The mobile app for Falling Fruit (fallingfruit.org)
  </description>
  <author email="info@fallingfruit.org" href="https://fallingfruit.org">
    Falling Fruit
  </author>
  <platform name="android">
    <preference name="android-minSdkVersion" value="21" />
    <preference name="android-targetSdkVersion" value="35" />
  </platform>
  <platform name="ios">
    <preference name="deployment-target" value="9.3.5" />
    <preference name="target-device" value="universal" />
  </platform>
</widget>
```

wbazant:
Closes #958 
Closes #977

### DIFF
--- a/ios/App/Podfile.lock
+++ b/ios/App/Podfile.lock
@@ -1,6 +1,8 @@
 PODS:
   - Capacitor (7.4.4):
     - CapacitorCordova
+  - CapacitorApp (7.1.2):
+    - Capacitor
   - CapacitorCamera (7.0.2):
     - Capacitor
   - CapacitorCordova (7.4.4)
@@ -17,6 +19,7 @@ PODS:
 
 DEPENDENCIES:
   - "Capacitor (from `../../node_modules/@capacitor/ios`)"
+  - "CapacitorApp (from `../../node_modules/@capacitor/app`)"
   - "CapacitorCamera (from `../../node_modules/@capacitor/camera`)"
   - "CapacitorCordova (from `../../node_modules/@capacitor/ios`)"
   - "CapacitorDevice (from `../../node_modules/@capacitor/device`)"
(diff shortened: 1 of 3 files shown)


wbazant:
I was just confused about it too! I long pressed on the app icon then the
info icon to open app info. Under Defaults i saw Set As Default, and a
toggle for the links (default was off). Then i started a new Firefox tab,
and pasted the share link.

Hm, a verified link will hopefully start with a default ask or default on
otherwise no one will use the affordance!

ezwelty:
@wbazant I have the Android version working (note: requires app to be signed with secret keystore). The iOS version sort of works (requires app to be signed via Apple Developer Falling Fruit team account): it recognizes that fallingfruit.org links can be opened in the app, but when I click the link, the app just opens. There is no rerouting to the specific URL. No idea why.

Now also closes #977.

wbazant:
I still had to open permissions and allow the links with my old app dev setup. Looks like mine is not the proper setup - my laptop has barely enough RAM to run it anyway so I won't fiddle with starting adb as a server or configuring npx cap to know about my Android Studio or whatever is missing. Thanks for documenting the release! I've also found a bug #978 but this PR looks good (for as much as I can check it) so I'll merge it in.

ezwelty:
@wbazant The app has to be signed and currently only I can do that because I'm the only one in possession of the keystore whose sha256 fingerprint is identified in the `./well-known/assetlinks.json`.

PR #985, opened 2026-02-19
Complete the two way binding between URL and map for location routes

### LINKED ISSUE #979: From user back to the location loses the view
Go to https://falling-fruit-web.pages.dev/locations/2291356, click on the user, then back: the map doesn't have a view and the URL does not sync.

wbazant:
Closes #979 

Works differently on desktop (map unmounts and remounts) and mobile (map component stays mounted, the map thinks it's 0px tall x 0px wide but still has the right center and zoom).


ezwelty:
@wbazant I confirm the particular issue is fixed. Are there other navigation sequences that you think could also be impacted that we should test here?

PR #993, opened 2026-02-21
Track being offline / back online

### LINKED ISSUE #975: Try to reconnect when briefly offline
I was out on a walk in a place where I think there's a gap in the Internet. I had a filter on to only show me the pear trees. Then I went to the settings to turn the labels on, back to the map, and the locations were gone. Moving the map didn't restore them.

Maybe a bug that produces an uncaught exception? Maybe there's no resilience on a failed nonfatal connection? I think we want to allow no internet for a bit, when on the map ( happens all the time in the field ) but there should be a message about it and an attempt to reconnect. We handled fatal failures in #900.

Especially important for the app, where 'refreshing the page' is an advanced user move - you have to go to info, force stop, and restart.

wbazant:
Closes #975

To test, be in the map view, set the browser to offline, and:
a) move the map
b) open filter counts
c) switch to list

d) Refresh, be in the list view, set the browser to offline, and open filter counts.

### DIFF
--- a/src/components/connect/ConnectLocation.js
+++ b/src/components/connect/ConnectLocation.js
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react'
+import { useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useDispatch, useSelector } from 'react-redux'
 import { toast } from 'react-toastify'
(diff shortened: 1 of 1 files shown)


ezwelty (dismissed):
@wbazant Thanks, the new behavior seems good! Also checked that it works in desktop and embed views.

* Switching to list view offline with a low zoom results in this:

image

Not necessarily a bug though, since in the mobile app I presume the image would be on the device and show up. Thoughts?

* As I wrote in Slack, "Results unavailable" in the list and filter view could be interpreted to mean that there are no locations on the map at this position. I wondered whether it should say something like "You are offline" or "Not available offline".
* Probably for another issue: During testing, I discovered that when signed in, submitting forms offline results in a error toast with network error. But when not signed in, recaptcha throws an error that isn't being caught, and the form stays stuck in submitting indefinitely.

Issue #997, opened 2026-02-27
Hide location delete by default to avoid flashing visibility


ezwelty:
To reproduce, go to map in desktop mode and click on locations that aren't your own (and thus cannot delete):

https://github.com/user-attachments/assets/4313a1ab-2c2a-4aff-b35a-53cb4825863a

PR #998, opened 2026-02-27
 Fix delete button flashing visibility 

Abmarne:
Summary
fix: #997
Fixed the delete button briefly flashing visible when viewing other users' locations in desktop mode
Problem
When clicking on a location that doesn't belong to the current user, the delete button would briefly appear before being hidden. This happened because when locationData was null (during loading), the comparison null?.user_id === undefined evaluated to true (both are undefined), incorrectly showing the delete button.

Solution
Added a check to ensure locationData exists before comparing user IDs:

Before: locationData?.user_id === user?.id
After: locationData && locationData.user_id === user?.id
This ensures the delete button remains hidden until the location data has fully loaded and the ownership can be properly determined.

Testing
Go to map in desktop mode
Click on a location that isn't your own
Verify the delete button is hidden (not flashing)

### DIFF
--- a/src/components/entry/useDeleteLocation.js
+++ b/src/components/entry/useDeleteLocation.js
@@ -35,7 +35,7 @@ export const useDeleteLocation = () => {
     toast.info(t('info_message.cannot_delete_location_with_other_reviews'))
   }
 
-  const isDeleteVisible = locationData && locationData.user_id === user?.id
+  const isDeleteVisible = locationData?.user_id === user?.id
   const isDeleteDisabled =
     reviews?.some((review) => review.user_id !== user?.id) || false
 


src/components/entry/useDeleteLocation.js
```diff
@@ -35,7 +35,7 @@ export const useDeleteLocation = () => {
     toast.info(t('info_message.cannot_delete_location_with_other_reviews'))
   }
 
-  const isDeleteVisible = locationData?.user_id === user?.id
+  const isDeleteVisible = locationData && locationData.user_id === user?.id
```
wbazant: I think this is the right fix! I'll check it tomorrow.

PR #1008, opened 2026-03-02
More marker updates in LocationMarkers

### LINKED ISSUE #999: Location marker does not move on position edit
Location marker does not move when the position is edited. I presume this regression was introduced in the switch to optimized markers.

Refreshing or moving the map updates the marker.

### LINKED ISSUE #1000: Location labels do not update on language change
Location labels do not update on language change. Reloading or panning the map to and fro so that they reload updated the labels.

Related to #999 in that I suspect it was introduced in the switch to optimized markers.

@wbazant I presume you looked into this at the time and it's not trivial to reload the map automatically on language change rather than asking the user to reload the page?

wbazant:
Closes #999
Closes #1000

### DIFF
--- a/src/components/map/LocationMarkers.js
+++ b/src/components/map/LocationMarkers.js
@@ -184,6 +184,14 @@ const createLabel = (
     this.draw()
   }
 
+  label.updateHtml = function (newHtml) {
+    if (!this.div || this.labelHtml === newHtml) {
+      return
+    }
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant Tested adding, deleting, and moving a location. Also tested changing language. The issues appear fixed. Thanks!

PR #1011, opened 2026-03-11
Fail when captcha fails

### LINKED ISSUE #1003: If recaptcha fails, the form gets stuck
Open a private browser tab, then location form without signing in, go offline, and try to submit the form. If you're not signed in, a recaptcha will appear, I got : "Cannot contact reCAPTCHA. Check your connection and try again. " . The form stays stuck in submitting indefinitely.

wbazant:
Closes #1003 

It looks like there's no way to recover from captcha failures (the API allows a reset, but it doesn't reliably work). Treat the error as fatal by going to /error/fatal. I'm preserving the state of the form, putting it in the URL, and repopulating on the way back, so at least a Reload will restore a description etc. that has been typed in.

Also, try to avoid failures in the first place by checking navigator.onLine - captcha won't work if this is false.

Easiest to test the commits separately - open a form in a private tab, fill in, set to offline, and submit. This should go to the error page in the first commit, and show a toast afterwards.

### DIFF
--- a/public/locales/ar.json
+++ b/public/locales/ar.json
@@ -65,6 +65,7 @@
       "not_supported": "تحديد الموقع الجغرافي غير مدعوم"
     },
     "language_changed_refresh_to_reload_map": "تم تغيير اللغة، يرجى تحديث الصفحة لإعادة تحميل الخريطة",
+    "recaptcha_failed": "فشل التحقق من ReCAPTCHA",
     "results_unavailable": "النتائج غير متاحة",
     "unknown_error": "خطأ غير معروف"
   },
--- a/public/locales/de.json
+++ b/public/locales/de.json
@@ -65,6 +65,7 @@
       "not_supported": "Geolokalisierung wird nicht unterstützt"
     },
     "language_changed_refresh_to_reload_map": "Sprache geändert. Bitte lade die Seite neu, um die Karte zu aktualisieren",
+    "recaptcha_failed": "ReCAPTCHA fehlgeschlagen",
     "results_unavailable": "Ergebnisse nicht verfügbar",
     "unknown_error": "Unbekannter Fehler"
   },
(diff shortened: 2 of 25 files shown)


ezwelty:
@wbazant I tested all the forms where recaptcha comes into play, mostly on the first commit in order to have more control when testing.

* [x] Add location and review
* [x] Edit location
* [ ] Review location – Caught with toast. If not, form state restored except for photos.
* [ ] Report location – Caught with toast. If not, form state lost. Reloads to view location page. Opening this modal form while offline is enough to trigger the fatal error page, regardless of commit.
* [ ] Submit new type – Caught with toast. Form state lost and add/edit location form state lost. Reloads to add/edit location page. Opening this modal form while offline is enough to trigger the fatal error page, regardless of commit.
* [x] Sign up – Caught with toast ("Sign up failed: Network Error"). Form state preserved.
* [x] Reset password – Caught with toast ("Sign up failed: Network Error"). Form state preserved.
* [x] Resend confirmation instructions – Caught with toast ("Sign up failed: Network Error"). Form state preserved.

Why do the last three forms already behave as desired? They always (?) show a recaptcha checkbox, cannot be submitted without the recaptcha being checked, and if checked and then switched to offline and submitted, fail gracefully with a toast and no need to reload the form. Can we just make all the other forms behave like this? It seems easier to work with and more transparent to the user to show them the recaptcha checkbox.

During testing, I occasionally saw the following, but difficult to reproduce.

> Uncaught (in promise) Error: reCAPTCHA Timeout at recaptcha__en.js:475:127

wbazant:
Thanks, that's a much better resolution! When the captcha is part of the form, the regular formik validation can prevent submission unless filled out, and there's no 'shows for some, but not for others' confusion if everyone has to at least click the 'I'm not a robot' checkbox. Closing this PR in favour of #1015.

PR #1012, opened 2026-03-11
On geolocation failures on the home page, go to the map

### LINKED ISSUE #866: "Position unavailable: " when location off
Asking for geolocation with location turned off at my android produced this message. It's not right, and looks to the user like something's not working correctly with the site.

The plan here is to first reproduce the problem - needs an android phone and wifi, when you run the site locally you can see the URL:
```
Compiled successfully!

You can now view falling-fruit in the browser.

  Local:            http://localhost:3000
  On Your Network:  {THIS IS THE ADDRESS TO USE IN THE MOBILE BROWSER :)}

Note that the development build is not optimized.
To create a production build, use yarn build.
```

Then find in the code where we are producing the message, and see what the right fix is - maybe we'll be able to add a missing error message, or maybe just change 'Position unavailable: ' into 'Position unavailable'.

### LINKED ISSUE #1001: Explore the map: if position unavailable, go to the map anyway
This is Claire on a simulated iPhone which is given location permissions and then fails:

https://github.com/user-attachments/assets/bc214846-6f3d-4357-8b73-2bf047ad2df3

From the user's perspective, it's confusing - they want to explore the map - see also #866.

wbazant:
Closes #1001
Closes #866

### DIFF
--- a/src/components/auth/HomePage.js
+++ b/src/components/auth/HomePage.js
@@ -66,7 +66,11 @@ const HomePage = () => {
   )
 
   const handleExploreMap = () => {
-    if (lastMapView || geolocationState === GeolocationState.DENIED) {
+    if (
+      lastMapView ||
+      geolocationState === GeolocationState.DENIED ||
+      geolocationState === GeolocationState.FAILED
+    ) {
       history.push('/map')
     } else {
       dispatch(requestGeolocation())
@@ -77,7 +81,10 @@ const HomePage = () => {
     if (lastMapView) {
       return
     }
-    if (geolocationState === GeolocationState.DENIED) {
+    if (
+      geolocationState === GeolocationState.DENIED ||
+      geolocationState === GeolocationState.FAILED
+    ) {
       history.push('/map')
     } else if (geolocation?.latitude && geolocation?.longitude) {
       history.push(
--- a/src/components/connect/ConnectGeolocation.js
+++ b/src/components/connect/ConnectGeolocation.js
@@ -224,9 +224,21 @@ const ConnectGeolocationInner = () => {
     if (geolocation.loading) {
       // Still loading, do nothing
     } else if (geolocation.error) {
-      dispatch(geolocationError(geolocation.error))
+      if (
+        geolocationState !== GeolocationState.FAILED &&
+        geolocationState !== GeolocationState.DENIED
+      ) {
+        dispatch(geolocationError(geolocation.error))
+      }
     } else if (!geolocation.latitude || !geolocation.longitude) {
-      dispatch(geolocationError({ message: t('error_message.unknown_error') }))
+      if (
+        geolocationState !== GeolocationState.FAILED &&
+        geolocationState !== GeolocationState.DENIED
+      ) {
+        dispatch(
+          geolocationError({ message: t('error_message.unknown_error') }),
+        )
+      }
     } else if (!googleMap) {
       dispatch(geolocationReceived(geolocation))
     } else if (isMapMoving) {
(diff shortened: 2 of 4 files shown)


ezwelty:
@wbazant Thanks again!

* #866: Could not recreate the original issue but using your override confirmed that the error message is correctly formatted.
* #1001: Could not recreate the original issue on iOS simulator (I tried to set Location to "None", then allow anyway). On the branch, the behavior is the same (goes to map but shows error toast "Position unavailable") except that it shows a weirder message.

    > Position unavailable: The operation couldn't be completed. (kCLErrorDomain 0.)

 Once I ended up in a situation where the toast reappeared every time I panned the map. If I turn Location services off in settings while the app is open on the map, then click on tracking icon, it spins forever. But I wonder whether we should test on a real iOS device before coming to any major conclusions...

@eclaires What steps did you use to originally reproduce #1001?

wbazant:
> Once I ended up in a situation where the toast reappeared every time I panned the map. 

I've found that one! The mechanism to change state from following to dot on after map pan accidentally allowed failed -> dot on -> failed ... leading to all those toasts.

I've also added code to remove the geolocation dot if the API returns a failure, in intermittent failures it gets brought back afterwards. Not sure if we're solving problems that happen outside mocks / simulators, seems most reasonable to merge the fixes but wait for reports from the field before messing with it more.

I could trigger the conditions for the bug with an elaborate geolocation mock that succeeds three times and then fails:
```
const originalWatchPosition = navigator.geolocation.watchPosition.bind(navigator.geolocation);
const originalClearWatch = navigator.geolocation.clearWatch.bind(navigator.geolocation);

const watchers = new Map();

navigator.geolocation.watchPosition = function(success, error, options) {
  console.log("[geo-mock] watchPosition called");

  let successCount = 0;

  const id = originalWatchPosition(
    (position) => {
      console.log("[geo-mock] real position received", position);

      if (successCount < 2) {
        successCount++;
        console.log(`[geo-mock] passing success #${successCount}`);
        success(position);
      } else {
        console.log("[geo-mock] stopping watcher and forcing error");

        originalClearWatch(id);

        if (error) {
          error({
            code: 2,
            message: "Simulated failure after 2 successes"
          });
        }
      }
    },
    (err) => {
      console.log("[geo-mock] real error received", err);
      if (error) error(err);
    },
    options
  );

  watchers.set(id, true);

  console.log("[geo-mock] watch id:", id);
  return id;
};

navigator.geolocation.clearWatch = function(id) {
  console.log("[geo-mock] clearWatch called for id:", id);
  watchers.delete(id);
  return originalClearWatch(id);
};
```

Also this one that fails intermittently:
```
const originalWatchPosition = navigator.geolocation.watchPosition.bind(navigator.geolocation);
const originalClearWatch = navigator.geolocation.clearWatch.bind(navigator.geolocation);

navigator.geolocation.watchPosition = function(success, error, options) {
  console.log("[geo-mock] watchPosition called");

  let count = 0;

  const id = originalWatchPosition(
    (position) => {
      count++;

      const cycleIndex = (count - 1) % 10; // 0–9
      const isSuccessPhase = cycleIndex < 3; // first 3 = success, next 7 = error

      console.log(
        `[geo-mock] event #${count} (cycle index ${cycleIndex}) → ${
          isSuccessPhase ? "SUCCESS" : "ERROR"
        }`
      );

      if (isSuccessPhase) {
        console.log("[geo-mock] passing real position", position);
        success(position);
      } else {
        console.log("[geo-mock] emitting simulated error");
        if (error) {
          error({
            code: 2,
            message: "Simulated intermittent failure (3 success / 7 error cycle)"
          });
        }
      }
    },
    (err) => {
      console.log("[geo-mock] real error received", err);
      if (error) error(err);
    },
    options
  );

  console.log("[geo-mock] watch id:", id);
  return id;
};

navigator.geolocation.clearWatch = function(id) {
  console.log("[geo-mock] clearWatch called for id:", id);
  return originalClearWatch(id);
};
```

PR #1015, opened 2026-03-19
Use formik recaptcha everywhere

### LINKED ISSUE #1003: If recaptcha fails, the form gets stuck

wbazant:
Closes #1003


ezwelty:
@wbazant Thanks, this seems like a much more robust fix. One thing I noticed during testing is that on narrow screens (e.g. iPhone SE) recaptcha appear cut at the right and bottom edges:

Screenshot 2026-04-09 at 23 27 27

What I wasn't able to test, because I was never challenged beyond clicking the checkbox, is how the challenges fit on narrow screens.

wbazant:
Thanks for catching that! I added a scaling of the component containing the captcha if the available width is less than recaptcha width. Somewhat annoyingly, the value is not right For modals, I tweaked the value until it looked good - there's an 'auto' margin on the modal so I couldn't find what the actually computed value is. I also tried a 'compact' recaptcha style, but that displayed the challenge to the right instead of above and frequently went out the screen, so scaling the component seems like the right hack. I get the challenge reliably when in private mode on firefox.

PR #1022, opened 2026-03-29
Saved locations

### LINKED ISSUE #856: Allow private tags / groups
From user feedback:
> Hey everyone, love the site!
> I just used it to get these black raspberries so thank you! Is there any way to save or tag a location? I see a couple I want to hit in the next couple days but I'd like to notate or tag them somehow so they standout from other ones.

We don't quite support it through the UI but had some offers to the user:
- add a review so it shows up on activity feed
- add a route on the legacy site
- add browser bookmarks

Related to #725 , as another 'private data' use case.

### LINKED ISSUE #1031: Password set page misses a bit of verbiage
We had:
```
users.change_password
users.change_your_password
```
With the second one missing from English. Found by the translation script checking translations which I don't often do.

I must have deleted it from one place, and not from everywhere. Issue is solved by switching to users.change_password. Nice to still find an obvious bug!

wbazant:
Closes #856 
Closes #1031


ezwelty:
@wbazant Thanks! I took it for a spin. Does what we need! I just have some ideas for light fine tuning:

Location page

* I suspect a + or bookmark with + would more clearly communicate the action
* It would be nice if a saved location looked more different on the location page than just the text changing from "Save" to "Saved". e.g. An empty icon or button becomes filled?
* The button order may be worth changing. Perhaps [ Save ] first, as outline turning to filled on save, [ Review ] as filled button (as now), and [ Report (least common) ] off to the side with either outline (as now) or only icon and text but no outline?

Location list page

* Clicking on a location in a list brings up a menu "View on map" "Remove from list". I think it should just link straight to the map, dropping the list. Remove is already available with the 'x'.
* Adding a new list closes any lists with locations loaded. Can we get rid of this behavior? I would suggest moving the dropdown action to the list name, move "X locations" to the top cell below the list name, and default all lists open showing locations.
* It feels like there are too many lines of the same color and thickness. Can the lists be slightly more visually separated from each other, and perhaps fewer lines used in styling each list?

wbazant:
Thanks, will have a go! For the button order, I think I will do:
``` 
[ Review ]  [ Save ]                      [ Report ]             ( logged in )
[ Review ]                                [ Report ]             (logged out)
````

The original thinking was to make the save appear without messing anything up, but reviewing and saving is more similar and of higher emphasis  than report. Maybe I'll try ` [ Saved ] ` getting primary styling and ` [ Save ] ` keeping secondary styling, so it's noticeably different, and the colour change to primary also happens on the location dot and in the list as indicators of being saved.

ezwelty:
@wbazant Two bugs (or just that you haven't gotten this far yet to wiring things up) preventing further testing:

* [x] The API endpoint is being called without an authorization header, hence error
* [x] The API endpoint is being called with `embed[]=locations`. This should be `embed=locations`.

Also, I think `/lists` should be changed to `/users/:id/lists` (to match activity) or `/account/lists`.

wbazant:
@ezwelty ready to review!

Main changes were in dealing with the 'loading' gap everywhere and syncing changes around: when I add / delete / rename, I patch the data instead of refetching, and updates in redux save slice are propagated to the list, map, and location slice for being in list vs not.

Things in the API that I'm working around or not using:
1. The location ships with lists, which I can only use for determining whether the location is in some list or not, I refetch all the lists anyway. I could get in_list as a boolean 
2. I don't use descriptions (except I have to pass a null one when renaming a list)
3. I don't use created / updated date for lists, they're probably practical to have in the database anyway

Also changed a saved locations page layout into a less angular / tabley one. Still pending at some point, go to a list-location and allow going 'back' - a contributor is working on a similar story for going back from activity page so I'll let them have a go at it first.

ezwelty:
@wbazant Great progress, sorry for the long wait. I think I found a bug:

* [x] Open a location, click Save, add a new list, then select the new list. The list name goes gray and I get the spinning cursor when hovering over it. The Save button does not go to "Saved", the map icon does not turn orange, ... Open another location, click Save, and the new list is shown, with gray text, and highlighted in orange despite the location not being in the list.

And more of a feature request which I think will be simple to implement:

* [x] Draw saved location markers (and their labels) above (with a higher z-index) so that they are visible above the sea of blue dots.

> The location ships with lists, which I can only use for determining whether the location is in some list or not, I refetch all the lists anyway. I could get in_list as a boolean

I see, because you need a `id` and `name` for all lists, plus whether the location is in that list (`in_list`). I could modify `location.lists` to that format. But if the list data in redux is kept up to date, who not just use that instead? If you're worried about `location.lists` being stale (it is current only as of `GET /locations/{id}`), I could add `GET /locations/{id}/lists` to request only `location.lists`.

> I don't use descriptions (except I have to pass a null one when renaming a list)
> I don't use created / updated date for lists, they're probably practical to have in the database anyway

That's fine. I still think these would be nice to add eventually to the list design, especially once lists can be shared (and thus necessarily have their own URL/page).

wbazant:
Thanks, that was a bug for adding a location to a just created list - an exception in redux. It was easier in localhost, because an exception was available. I added zIndex manipulation for saved markers and labels, so saved shows over non-saved. We also have a hover effect for moving the label of a hovered location to a higher pane, which became very weird when I tried to also apply it to markers, so hovered nonsaved marker doesn't show over saved one, only its label. 

> I could add GET /locations/{id}/lists to request only location.lists.

That would work and produce a small saving - in_list boolean on a location, and on click save, make a call to fetch the location's lists instead of all lists - maybe if we need to save on the back-end queries for this feature one day!

PR #1027, opened 2026-04-09
Japanese translation

wbazant:


ezwelty:
@wbazant Added Japanese common names.

PR #1028, opened 2026-04-11
Update Chinese translations for consistency

yanshili645:
您 is more appropriate and keeps consistent with other sentences.


wbazant:
Let's merge, would it make sense to also make the same change in the traditional character set?

In a few languages that I vaguely speak, I made the change for the informal version (Du instead of Sie in German, etc.) as it seems to match the tone of the site more. In Chinese, I have no feel if 你 → 您 is rightfully polite instead of too formal!

yanshili645:
If we treat users as a customer, 您 is more polite and respectful. If we see
users as a friend then we should use 你 in a more casual way. Which one we
should pick depends on how we position the application. So what’s your
preference?

wbazant:
Do they both appear online often, depending on the site? If yes then 你 is probably better for us, our user is a friend rather than customer, and probably young.

yanshili645:
> Do they both appear online often, depending on the site? If yes then 你 is probably better for us, our user is a friend rather than customer, and probably young.

Yes. So in this case we should use 你 everywhere as a UX principle. I'll try to find a session to go through them because I also find some awkward expressions in the current translation.

PR #1029, opened 2026-04-12
Show users with empty name as number

### LINKED ISSUE #1021: Support users with empty name
We're migrating away from 'if name is empty, user wishes to be anonymous' and towards 'you're either logged in or logged out'. I'm currently a 'no name' account to test this.

1. let author field be legacy only (api)
This still says added by Wojtek: https://beta.fallingfruit.org/locations/2112808/@55.8268479,-4.2613612,17z
We could introduce an extra field, `user_name`, that comes from a join or is batch updated when a user changes their name, and let `author` be an old legacy field that stands for the user name at the time of location add. Then only show author when user id is missing - or we could also not overthink the schema, and give `author` the semantics of a user name, with old values as exception.

2. Display authorship for users with empty name
This doesn't have author field but has user id: https://beta.fallingfruit.org/locations/2292033/@55.8296021,-4.2620915,19z 

3. Remove the 'user prefers to remain anonymous' from user profile endpoint (api)
This should load: https://beta.fallingfruit.org/users/57182 . Note, the location 2112808 (legacy author field plus user ID) links to the profile, that then isn't displayable

4. Remove the 'user prefers to remain anonymous' from the changes endpoint (api)
https://beta.fallingfruit.org/users/57182/activity should load - notably when I'm logged in it doesn't currently let me view my profile

wbazant:
Closes #1021 

Changes to user profiles, activity page, and location overview: show users with empty name as a number.

@ezwelty are we aligned on having this instead of the private setting on profiles? I'm also not sure how 'private' setting works now, for example 57182 is a user with empty name who isn't private and their profile can be viewed, but the activity page now links to many private users.

### DIFF
--- a/src/components/activity/DiaryEntry.js
+++ b/src/components/activity/DiaryEntry.js
@@ -127,6 +127,8 @@ const ActivityTextComponent = ({
   const isRTL = i18n.dir() === 'rtl'
   const isCurrentUser = currentUserId && authorUserId === currentUserId
 
+  const displayName = author || (authorUserId ? `#${authorUserId}` : null)
+
   const LocationDisplay = () => (
     <>
       {hasLocationInfo ? (
@@ -151,14 +153,14 @@ const ActivityTextComponent = ({
       })}
       {isRTL ? (
         <>
-          {author && !isCurrentUser && (
+          {displayName && !isCurrentUser && (
             <>
               {authorUserId ? (
                 <AuthorLink to={`/users/${authorUserId}`} onClick={onClickLink}>
-                  {author}
+                  {displayName}
                 </AuthorLink>
               ) : (
-                <span>{author}</span>
+                <span>{displayName}</span>
               )}
               {' — '}
             </>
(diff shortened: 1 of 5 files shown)


ezwelty:
@wbazant We are aligned, but we've now entered an awkward transition phase where the API is lagging behind in production and I'm the only one able to run API in development. In the dev version of the API, there are no private profiles and in nearly all cases (there are only 11 users with a null name and one or more locations or reviews), it would return the user's current name alongside the user's id.

wbazant:
Sure, then think this change is fine to review / merge ! https://issue-1021.falling-fruit-web.pages.dev/changes links to quite a few profiles that can't be viewed, but if that is going away as part of the switchover, it's not a problem. The user profile can be tested on https://issue-1021.falling-fruit-web.pages.dev/users/57182.

ezwelty:
@wbazant Found two bugs

* [x] The user id is missing from the activity page : https://issue-1021.falling-fruit-web.pages.dev/users/57182/activity
* [x] The link to the user id is missing from reviews with `user_id` but null `author` : https://issue-1021.falling-fruit-web.pages.dev/locations/2112021

Issue #982, opened 2026-02-19
Make it clearer that location position is editable


wbazant:
This is the edit location form:

https://falling-fruit-web.pages.dev/locations/2291356/edit/@41.2940872,-81.3199479,16z

<img width="717" height="945" alt="Image" src="https://github.com/user-attachments/assets/f324fa27-fa77-4024-bb23-648fb6b312d7" />

A user didn't know they can actually click the orange link and edit the position. I wonder if we can make it clearer? Some ideas (that I'm not entirely happy with) are pencil icon, an underline, and a button that says 'adjust position'.

ezwelty:
@wbazant Pencil icon after the coordinates seems reasonable. Or replacing the marker icon with the pencil icon? Or outlining the marker icon and coordinates such that it looks like a form input like the rest of the form?

Thirukumaran-T:
@wbazant is the issue still open.. shall i work on it

wbazant:
@Thirukumaran-T sure, please do! Let's try replacing the marker icon with the pencil icon when the position is editable.

Thirukumaran-T:
<img width="426" height="702" alt="Image" src="https://github.com/user-attachments/assets/fe23d537-af59-43ab-9b9a-6228cef2acf6" />
@wbazant When the position is editable!, it seems disabled when i try for editing the locations

wbazant:
@Thirukumaran-T  have a look at what this looks like between desktop and mobile layouts. On desktop, we don't want a change, but on mobile we do. To see the mobile layout in your browser, make a window below 767 px in width - I open the console or drag it to make it smaller

Thirukumaran-T:
How about this @wbazant 

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/6e2b2018-b0c7-43f5-9bfb-0ab3df48d501" />

wbazant:
Thanks, that should be quite clear! Can we also try a custom icon that looks like a position marker with a pencil across it? Should be obtainable by giving two svg sources of position and pencil to a chatbot, or superimposing two icons on top of one another.

Thirukumaran-T:
Sure i'll try that..

Thirukumaran-T:
@wbazant How about this!

<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/82c59920-3457-4024-8ec3-2a51570c491e" />

wbazant:
[...]

PR #1032, opened 2026-04-21
Adding pencil icon for location edit

Thirukumaran-T:
It is fix for issue #982 

I have combined marker icon with location icon

It is tested and i'll share the image below

location with pencil

### DIFF
--- a/src/components/form/LocationForm.js
+++ b/src/components/form/LocationForm.js
@@ -1,4 +1,3 @@
-import { Pencil } from '@styled-icons/boxicons-solid'
 import { ErrorMessage, Form, Formik, useFormikContext } from 'formik'
 import { useEffect } from 'react'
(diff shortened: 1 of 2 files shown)


wbazant:
I checked out the branch locally and it looks nice on mobile, I think the subscript works quite well and it suggests being editable.

Can we have the original Map icon on desktop? The pencil there is not as good because it's not clickable. Make a change like 
```
import { useIsDesktop } from '../../utils/useBreakpoint'
...
const isDesktop = useIsDesktop()

...
isDesktop ? Map : PositionEditIcon
```

Thirukumaran-T:
Sure i'll do that changes

Thirukumaran-T:
I have updated the changes .. please review @wbazant

wbazant:
I checked it out locally again, and the change looks good! Thanks for implementing.

PR #1037, opened 2026-04-24
Fetch earliest, all, and more changes. Mark stale on delete

### LINKED ISSUE #1007: Recent activity page might not include very newest recent activity
Keep the recent activity page in sync with the data. 

Proposed:
- when revisiting the page through the main link, fetch the changes from current date to (last location loaded plus one second) and show a skeleton line in the top diary entry
- don't fetch when coming back from the location through the 'back' link
- only fetch more changes when scrolling to the bottom (we fetch more every time now - fix bug)
- when deleting location or review, mark the data as stale, and refetch when returning to the page (presence of last visited section is needed to make the back button work)
- when deleting, editing, adding location or review, mark user's own activity as stale, and make sure it refetches when coming back to own activity page

wbazant:
Closes #1007

### DIFF
--- a/src/components/activity/DiaryEntry.js
+++ b/src/components/activity/DiaryEntry.js
@@ -1,6 +1,7 @@
 import { MapAlt as Map } from '@styled-icons/boxicons-regular'
 import React, { useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
+import Skeleton from 'react-loading-skeleton'
 import { useDispatch, useSelector } from 'react-redux'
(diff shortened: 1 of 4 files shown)


ezwelty:
@wbazant Played around and seems to work! One thing I noticed, peripherally related, is that the built-in back button from the user's activity page goes back to the account page, not the map, but I think I read somewhere that this was planned to be changed?

wbazant:
Thanks! I think you've found a bug in the desktop layout, I made an issue for it.

PR #1039, opened 2026-04-27
don't call syncViewToBrowserUrl from onChange handler after browser b…

### LINKED ISSUE #991: Pressing back URL only goes back every other time
Go to https://falling-fruit-web.pages.dev/map/@41.2940872,-81.3199479,16z, and pan the map down two times. Check that the URL changes every time. Then press the back button four times, this is what I see:
- move back
- stay
- move back
- stay

Navigating Falling Fruit through the back button is quite niche, especially for map pans, so this is a low severity bug. Might be a fun challenge to find! Can't use AI for this one. adding console.trace in history.push could be helpful to solve.

badaminatasha:
…ack button navigation

closes https://github.com/falling-fruit/falling-fruit-web/issues/991

## Problem
- Back button causes map to pan -> triggers GoogleMapReact onChange event -> invokes syncViewToBrowserUrl which pushes the url that was just popped back into browser history -> duplicated entry in browser history -> user must hit browser back button twice for it to actually go back
- In some browsers (tested edge and chrome), back button seems to break after one use (stops going back or goes all the way back to new tab page). I'm convinced this is due to [user activation requirements](https://developer.mozilla.org/en-US/docs/Web/API/Window/popstate_event#when_popstate_is_sent): syncViewToBrowserUrl pushes to history while/immediately after popstate, maybe browser flags it and considers it not preceded by user interaction with the page.

Ideally syncViewToBrowserUrl is only called after user interaction event. maps api [does expose events for user interaction](https://developers.google.com/maps/documentation/javascript/events#EventsOverview), but not for user-initiated zoom/scroll.

## Partial solution
- set "back button pressed since last url sync" flag to true on popstate
- MapPage.js next attempt to sync url is ignored and the flag is set back to false
- only user-initiated map manipulations are synced to browser url

## Issues
Two race conditions, both arguably minor:
- back button -> valid map bounds-changing user input in rapid succession
- valid map bounds-changing user input -> back button in rapid succession
In both cases, the issue is that the popstate event fires before the map pan from valid user input finishes. As a result, the flag is set to true (skip sync) before onChange is fired, so the url isn't synced for one input.

### DIFF
--- a/src/components/map/MapPage.js
+++ b/src/components/map/MapPage.js
@@ -142,14 +142,14 @@ const clusterBounds = ({ lat, lng, zoom }) => {
     east: bounds.east * (360 / EARTH_CIRCUMFERENCE),
   }
 }
-const makeHandleViewChange = (dispatch, googleMap, history) => {
+const makeHandleViewChange = (dispatch, googleMap, history, setPopstate) => {
   const throttledDispatches = throttle((newView) => {
     dispatch(updateLastMapView(newView))
     dispatch(fetchLocations())
     dispatch(fetchFilterCounts())
   }, 1000)
 
-  return (googleMapReactCallbackArg) => {
+  return (popstateSinceLastSync) => (googleMapReactCallbackArg) => {
     const center = googleMap.getCenter()
     const newView = {
       center: { lat: center.lat(), lng: center.lng() },
@@ -159,8 +159,13 @@ const makeHandleViewChange = (dispatch, googleMap, history) => {
       height: googleMap.getDiv().offsetHeight,
     }
     throttledDispatches(newView)
-    if (googleMapReactCallbackArg) {
+
+    // Don't sync if bounds changed due to browser back button
+    if (!popstateSinceLastSync && googleMapReactCallbackArg) {
       history.syncViewToBrowserUrl(newView)
+    } else if (popstateSinceLastSync) {
+      // reset after first skip
+      setPopstate(false)
     }
   }
 }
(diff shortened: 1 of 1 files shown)


wbazant:
@badaminatasha I have an idea for a simpler patch - I moved the map four times in a square, then pressed four times back, and got taken back four times:

```
diff --git a/src/utils/useAppHistory.js b/src/utils/useAppHistory.js
index 5fac3916..5a4ce8ae 100644
--- a/src/utils/useAppHistory.js
+++ b/src/utils/useAppHistory.js
@@ -34,8 +34,11 @@ export const useAppHistory = () => {
   }
 
   const syncViewToBrowserUrl = (newView) => {
-    const newUrl = currentPathWithView(newView)
-    window.history.pushState(null, '', newUrl)
+    const newPath = currentPathWithView(newView)
+    const oldPath = window.location.pathname
+    if (oldPath !== newPath) {
+      window.history.pushState(null, '', newPath)
+    }
   }

```

Do you think it works? We can still listen to popstate and maintain the flag if there's a corner case I can't think of, but maybe it's enough not to add the extra state entries.

badaminatasha:
That seems fine to me. The key thing is that the url sync should never be called during popstate. The only edge case for your solution I was thinking of is if it's somehow possible that the final state of the bounds is slightly different from the url old url upon hitting the back button, but in practice I couldn't trigger that while testing this. Did you want to merge your change then?

wbazant:
Superseded by https://github.com/falling-fruit/falling-fruit-web/pull/1041

PR #1045, opened 2026-05-05
Readjust narrow desktop layout

### LINKED ISSUE #980: Readjust the narrow desktop layout so there's no overflow after opening the menu
We have a 'narrow desktop layout' to catch the header being still desktop size, but not wide enough to show both the menu and the sign in / sign up buttons. I just noticed it's not quite right - the values can still overflow, the little triangle after "about" is squished and the social icons get cramped:

Image

I like the social icons and the logo staying to anchor the whole component in place, but they also take up space. So maybe don't show the social icons when there's no space for them? We can have a 'narrow desktop layout' split into a 'very narrow' and 'slightly wider' if it helps.

Thirukumaran-T:
Closes #980

### DIFF
--- a/src/components/desktop/Header.js
+++ b/src/components/desktop/Header.js
@@ -31,14 +31,39 @@ const AuthLinksList = styled.ul`
   gap: 12px;
 `
 
-const TIGHT_LAYOUT_MAX_WIDTH = 1080
-const TIGHT_LAYOUT_MAX_WIDTH_USER = 980
+const VERY_NARROW_MAX_WIDTH = 900
+const NARROW_MAX_WIDTH = 1080
+const NARROW_MAX_WIDTH_USER = 980
+
+function getLayoutType(user) {
+  const width = window.innerWidth
+
+  if (user) {
+    if (width <= VERY_NARROW_MAX_WIDTH) {
+      return 'very-narrow'
+    }
+    if (width <= NARROW_MAX_WIDTH_USER) {
+      return 'narrow'
+    }
+    return 'normal'
+  } else {
+    if (width <= VERY_NARROW_MAX_WIDTH) {
+      return 'very-narrow'
+    }
+    if (width <= NARROW_MAX_WIDTH) {
+      return 'narrow'
+    }
+    return 'normal'
+  }
+}
 
 function layoutIstight(user) {
-  return (
-    (user && window.innerWidth <= TIGHT_LAYOUT_MAX_WIDTH_USER) ||
-    (!user && window.innerWidth <= TIGHT_LAYOUT_MAX_WIDTH)
-  )
+  const layoutType = getLayoutType(user)
+  return layoutType === 'narrow' || layoutType === 'very-narrow'
+}
+
+function layoutIsVeryNarrow(user) {
+  return getLayoutType(user) === 'very-narrow'
 }
 
 const StyledHeader = styled.header`
@@ -120,8 +145,16 @@ const NavLi = styled.li`
   display: flex;
   justify-content: stretch;
   align-items: stretch;
-  min-width: ${() =>
-    window.innerWidth <= TIGHT_LAYOUT_MAX_WIDTH ? '80px' : '110px'};
+  min-width: ${() => {
+    const width = window.innerWidth
+    if (width <= VERY_NARROW_MAX_WIDTH) {
+      return '70px'
+    }
+    if (width <= NARROW_MAX_WIDTH) {
+      return '80px'
+    }
+    return '110px'
+  }};
   margin: 0;
   color: ${({ theme }) => theme.secondaryText};
   cursor: pointer;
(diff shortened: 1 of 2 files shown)


Thirukumaran-T:
Please review this PR! @wbazant ..

src/components/desktop/Header.js
```diff
@@ -250,6 +283,7 @@ const SignupButton = styled(Button)`
 const StyledSocialButtons = styled(SocialButtons)`
   margin: 0 5px;
   display: flex;
```
wbazant: Can we remove this line?

wbazant: The display is redefined below.

src/components/desktop/Header.js
```diff
@@ -120,8 +145,16 @@ const NavLi = styled.li`
...
   justify-content: stretch;
   align-items: stretch;
-  min-width: ${() =>
-    window.innerWidth <= TIGHT_LAYOUT_MAX_WIDTH ? '80px' : '110px'};
+  min-width: ${() => {
```
wbazant, line 156: Good opportunity to use the function you defined above, getLayoutType!

src/components/desktop/Header.js
```diff
@@ -250,6 +283,7 @@ const SignupButton = styled(Button)`
 const StyledSocialButtons = styled(SocialButtons)`
   margin: 0 5px;
   display: flex;
+  display: ${({ user }) => (layoutIsVeryNarrow(user) ? 'none' : 'flex')};
```
wbazant: I've not yet ran the code, but will this get hidden always or only if the menu is open? We want them available at least sometimes.

wbazant (changes requested):
Thanks, I think the approach will work! I reviewed the code and requested some small changes.A screencast of the header opening in various widths would help me review the resulting UI if its not hard for you to make.

Thirukumaran-T:
I'll do the required changes and update!

Thirukumaran-T:
I have updated the PR @wbazant .. please review and i'll upload the pictures
one
 below
two
three

wbazant:
Thankjs for making the changes! I'm back from holiday and I have my laptop back, so I checked out the branch and tried different widths. I think your 'very narrow' layout needs a fix - it hides the social buttons when they don't need hiding, and there's still an overflow:
[Screencast from 2026-05-07 21-01-55.webm](https://github.com/user-attachments/assets/c900d1b8-0653-4f7d-99da-2f126e271644)

Thirukumaran-T:
@wbazant  I have refacored the code for  very narrow layout and the social buttons are only hiding when truly necessary and the overflow.

Hope you had a nice holiday! 🌴.

wbazant:
This is not the right fix, I would like the social buttons to be hidden in this situation:

Screenshot from 2026-05-08 10-19-43

Thirukumaran-T:
@wbazant i have refactored the code for very narrow layout.. please review

wbazant:
No, that change does not work.

Thirukumaran-T:
Can you please suggest me a fix..while testing i have seen for a very narrow layout the social icons is not visible.. @wbazant  is it still overflowing!

wbazant:
Sure, happy to clarify what the problem is here! By the overflow I mean, the menu part of the ribbon doesn't fit in the width it has and the browser tries to squish it in, which looks like the little triangle that is usually to the right of the 'About' menu getting its own line. Your change doesn't address the problem, because you're removing the social buttons when the menu is not visible, instead of when it's visible. I think if you fix that by changing the logic a bit, remove the CSS changes you introduced while trying to solve the problem, and test again in your browser, the PR will be good to go.

Thirukumaran-T:
Okk!.. i'll do that required changes

Thirukumaran-T:
@wbazant can you review my changes once

wbazant:
I looked at the code change and it doesn't look like it addresses what's needed - we want the social buttons visible at the start and hidden once the menu button is pressed, so the menu content has enough space. Your change instead hides them at the start and makes them visible once the menu button is pressed, so the menu content doesn't gain any additional space. Also, sorry, I see you're going in circles on this and I'm not really sure how to unblock you - maybe starting again and making sure you can reproduce the issue in your browser before making any changes, or maybe I will take this issue later and you could do the new translation into Tamil instead?

Thirukumaran-T:
@wbazant i'll try it once.. if it is not resolved you can take this issue and i'll move to translation

Thirukumaran-T:
@wbazant have you reviewed it!..

wbazant:
@Thirukumaran-T I've not run it but I looked at the code diff, it has a bunch of changes, I imagine you kept prompting an LLM when I was telling you your changes don't work. I also realised we might actually just need two lines:

```
+ const VERY_NARROW_LAYOUT_WIDTH = ...
```

```
        <div className="auth-social">
          {!isMobileMenuOpen && <UserMenu />}
-           <StyledSocialButtons />
+        {(!isMobileMenuOpen || window.innerWidth > VERY_NARROW_LAYOUT_WIDTH ) &&  <StyledSocialButtons /> }
```

Feel free to try that on a fresh checkout and see if it hides the buttons when the menu is open, so the menu has enough space in a narrow layout? Can you experimentally find a good value for VERY_NARROW_LAYOUT_WIDTH?

Thirukumaran-T:
I have understood the issue but it seems overwork for it.. i have to learn more about this i'll move to tamil translation can you take over this issue and raise PR .. i'll learn from your PR what has to be really changed

wbazant:
Replaced by #1108.

PR #1052, opened 2026-05-08
Don't show back button for own activity page on desktop

### LINKED ISSUE #1046: Back from your activity on desktop
Ethan: One thing I noticed, peripherally related, is that the built-in back button from the user's activity page goes back to the account page, not the map

_Originally posted by @ezwelty in https://github.com/falling-fruit/falling-fruit-web/issues/1037#issuecomment-4388848014_
            

Let's solve it by not offering the back arrow for your activity page on desktop: still offer it on another user's activity and on mobile. The desktop user will be able to use a top ribbon to navigate away.

wbazant:
Closes #1046

### DIFF
--- a/src/components/account/AccountPage.js
+++ b/src/components/account/AccountPage.js
@@ -134,12 +134,16 @@ const AccountPage = () => {
                 <Link to="/account/change-password">
                   {t('users.change_password')}
                 </Link>
-                <Link to={`/users/${user.id}/activity`}>
-                  {t('users.your_activity')}
-                </Link>
-                <Link to="/account/lists">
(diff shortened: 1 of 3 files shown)


ezwelty:
@wbazant If that's the solution, then the same should be done to the saved locations page. But it feels odd that "Change email" and "Change password" have a back but "Your activity" and "Saved locations" don't. Should the latter two also be removed from the account links in desktop, since they have their own menu entry?

wbazant:
Sure, I also removed the back arrow from saved locations on desktop and removed the links from the location page.

PR #1054, opened 2026-05-08
If address is null, render lat and lng instead

### LINKED ISSUE #1053: Saved locations without address
Show a pin. Forgot to do it in original issue.

wbazant:
Closes #1053

Screenshot from 2026-05-08 12-06-29


ezwelty:
@wbazant Looks good, thanks!

PR #1055, opened 2026-05-08
Back from location

### LINKED ISSUE #1024: back to activity page on mobile
On desktop, the 'back' button on the location page correctly returns to the previous location based on redux state: either global activity page, or author's activity page, or map.

On mobile, we just go to the view of the location on the map, and there is no 'back'. We're about to add a feature where we'd also like to go back from a location page - saved lists - so it's time to solve this.

The links to location page on activity page and user activity page should go to the same view as location from list: full screen, with options to either go back or to the map. See the relevant code for implementation ideas. AI agents welcome :).

wbazant:
Closes #1024

### DIFF
--- a/src/components/account/SavedLocationsPage.js
+++ b/src/components/account/SavedLocationsPage.js
@@ -15,15 +15,18 @@ import { Link } from 'react-router-dom'
 import styled from 'styled-components/macro'
 
 import {
+  clearLastViewedListId,
   fetchLists,
   removeList,
   removeLocationFromList,
   renameList,
+  setLastViewedListId,
 } from '../../redux/saveSlice'
 import { BackButton } from '../ui/ActionButtons'
 import { theme } from '../ui/GlobalStyle'
 import Input from '../ui/Input'
 import { Page } from '../ui/PageTemplate'
+import ResetButton from '../ui/ResetButton'
 
 const ListCard = styled.div`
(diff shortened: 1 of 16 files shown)


ezwelty:
@wbazant The back button works as expected, but the map response has a few bugs.

First version:

1. Account > Your activity > Click on a location OR Account > Recent activity > Click on a location
2. Click map icon button
3. Map marker is visible, but other nearby locations are not shown!
4. Click on the map. Drawer closes and marker disappears!

Second version:

1. Account > Saved locations > Click on a location
2. Click map icon button
3. Map is empty of markers and of the previous map bounds!

wbazant:
@ezwelty thanks for catching that! the issue was to do with map bounds, for /list there's always a map with its corresponding view that just loaded, but not on activity or saved locations pages. First part of the solution was to specify the view as centered on the location.

After this, I was going to the right place but the mechanism in ConnectPath happened when the map wasn't on the screen, so the map was setting its bounds to a degenerate zero area square and fetchMapLocations wasn't getting the bounds.

The solution was to change the URLs so we first go to /locations-standalone/:locationId (was /locations/:locationId/standalone in the bad version, and /list-locations/:locationId on main) so we first set the state in redux before doing anything else, and recognising the fully open drawer in ConnectPath to delay the fetch.

PR #1058, opened 2026-05-11
Czech translation

### LINKED ISSUE #1049: Czech translation
Requested by email.

wbazant:
Closes #1049


ezwelty:
@wbazant Can't comment on the Czech, but why did `time.days` get moved to `pages.changes.days_ago` but remain equivalent? It isn't "days ago" because the "ago" part is still added by `time.time_ago` (`{{time}} ago`).

wbazant:
I intended to move just so the phrase is scoped to this one place, but didn't look beyond just renaming the key! I'll change the two keys to a new machine translated one. I've not heard back from Barbora about whether she's okay to be listed as a translator, could you maybe add common names and email her saying it's all ready and ask the same + any more edits?

ezwelty:
@wbazant Fine if you want to move the keys, but then I'd suggest moving all three and keeping their names. So `pages.changes.days`, `pages.changes.time_ago`, and `pages.changes.last_24_hours`.

I can followup with Barbora.

PR #1071, opened 2026-05-15
Common styling for forms

### LINKED ISSUE #925: Common styling for forms
We have three sources of styling, `src/components/form/FormLayout.js`, `src/components/auth/AuthWrappers.js`, and `src/components/ui/Modal.js`.

FormLayout.js is probably the best one, but needs extending to also work in a desktop without panes. Improvements that are in some places that we would like to see everywhere:
* [ ]  textareas are shorter on shorter screens
* [ ]  layout in em instead of pixels
* [ ] elements take up the whole space inside the container

Also:
* [ ]  common heights ( right now, name and type in Add New Type are of different heights, and password is slightly taller than email in Email Settings )
* [ ] buttons align to the end on a wide desktop form, and split the full width in side pane and mobile

See https://github.com/falling-fruit/falling-fruit-web/pull/922#issuecomment-3646566397.

wbazant:
Closes #925

### DIFF
--- a/src/components/account/AccountPage.js
+++ b/src/components/account/AccountPage.js
@@ -10,11 +10,11 @@ import { editProfile, logout } from '../../redux/authSlice'
 import { pathWithCurrentView } from '../../utils/appUrl'
 import { useAppHistory } from '../../utils/useAppHistory'
 import { useIsDesktop } from '../../utils/useBreakpoint'
-import { FormButtonWrapper, FormInputWrapper } from '../auth/AuthWrappers'
 import { Input, Textarea } from '../form/FormikWrappers'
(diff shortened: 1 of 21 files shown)


ezwelty:
@wbazant Looks good!

Something is strange though – the recaptcha is back to being too big and truncated, which I thought we fixed already recently, but cannot find where a regression would have been introduced.

wbazant:
I tried to reproduce, the recaptcha box proportionally shrinks when I make the browser window smaller / bigger, and then I had only one open recatptcha challenge - it came up full size and not completely on the screen. Also looked good on the app built off the main branch.

PR #1075, opened 2026-05-17
Revisit saving photos to device

### LINKED ISSUE #358: Save photos to mobile device
Currently, photos that are taken directly in the app are not saved onto the device, so they are effectively lost to the user. Photos taken in the app should also be saved onto the device, perhaps in a directory called "Falling Fruit" if not in the standard Camera directory. Is there a standard location for app-specific photos on Android and iOS?

ezwelty:
https://capacitorjs.com/docs/apis/camera#android suggests that `saveToGallery` requires `WRITE_EXTERNAL_STORAGE` on older Android versions, so I followed their recommendation of setting a `maxSdkVersion`.

I also:

* Switched to using the actual filename rather than an auto-generated one. Upon testing, the filename matches the one saved to the gallery, which is nice and feels like an upgrade, but does not match images chosen from the gallery, which is unfortunate. I also considered using `exif.DateTimeOriginal`, but on my phone it disturbingly does not follow ISO8601 (`2026:05:16 20:05:14`) so I tossed the idea as too risky.
* Set max size on both width and height, and moved JPEG quality and max size to constants. I do wonder whether it is appropriate to resize so aggressively, at least on the web version, and might suggest bumping from 1200 to 2048 pixels.

Something annoying but that we'll probably have to live with is that the image saved to the gallery is the reduced size and devoid of metadata, not the original as if it had been first taken with the camera.

... Oops: Those first three commits I had intended to push to main directly as part of routine dev maintenance.
... And https://github.com/falling-fruit/falling-fruit-web/pull/1075/commits/34a7a1be5df9468a0e6a5eca3583a67cfe13f748 meant for your #1058. My git-fu is rusty today.

Closes #358.


wbazant:
Thanks! I reproduced the bug, built the app from this branch, and got the photo saved into DCIM folder. I picked 1200 pixels with mobile internet away from big city in mind, but have no feel for what it does to the images, I think I just experimented with sizes and how they looked on my phone a bit. We can bump it up! It could be also a drop-down, compression: standard, HQ and full size, with standard as default.

PR #1079, opened 2026-05-18
beta release

### LINKED ISSUE #1068: `locations-standalone` break browser navigation
Introduced in a recent pull request, locations are linked from activity and saved locations pages as `locations-standalone`, which 

* https://beta.fallingfruit.org/users/1/activity > click >
* https://beta.fallingfruit.org/locations-standalone/2292559/@43.5185553,5.4635078,18z > redirect >
* https://beta.fallingfruit.org/locations/2292559/@43.5185553,5.4635078,18z

I'm sure there was a good reason for this, but it feels strange not to use the canonical locations URL (can a URL param be used instead?) and introduces buggy behavior in browser back/forward:

* Clicking back goes to: https://beta.fallingfruit.org/locations-standalone/2292559/@43.5185553,5.4635078,18z (so does nothing)
* Clicking back 3-5 times: https://beta.fallingfruit.org/users/1/activity (finally!)
* Clicking forward: https://beta.fallingfruit.org/locations/2292559/@43.5185553,5.4635078,18z
* Clicking back: blank tab!
* ...

edit: This issue also applies to locations from the location list

### LINKED ISSUE #1070: Changing language from settings crashes the app
Not reproducible in browser. Not yet tested on iOS. I presume this is related to the force-refresh on language change in the app, since app users don't have the option of "refreshing the page" to get the map to update. What's interesting is that it doesn't happen from the welcome page language dropdown, only from the language dropdown in the settings. Occurs whether signed in or not.

### LINKED ISSUE #1076: Saved locations page content is too narrow
The content on the saved location page is needlessly narrow (in web and mobile) and doesn't match other pages. Compare for example:

Image
Image

Probably all that is needed is removing some extra margins.

### LINKED ISSUE #1078: Adding new list should seed it with the location
A new list in the context of an open location should also add the location to the list - makes sense for what the user wants to achieve in that moment. Schedule the two API calls after each other.

wbazant:
Closes #1070 

@ezwelty how do you feel about an enhancement to drop `@` from the view syntax? I've confused myself with the test cases I used and gone in completely the wrong direction, wrote code to change how we store path segments, and then it didn't solve the issue. Easy to put it back in or remove but now that I look at the URL without `@`, I like it a bit more, and now is a great time to change URLs.

Closes #1068
I made the change to keep drawer state in params. Same commit because it also changes `src/utils/appUrl.js` a lot. It's definitely better - for example there's a skeleton loader shown when going from list to location - and we could maybe use params whenever there's a separate app view, like in the report modal.

On the other hand, it's harder to write transitions in this way. When keeping state in Redux, you can observe changes in state and make dispatches asynchronously, and have separate files for each individual tiny thing like e.g. a location marker on a map, but with url params, they start to clash with each other and make race conditions.

We have code for managing URLs to maintain view and params, so I added a new rule to it - a param with empty value is dropped - and whenever I navigate away from the context of panes I add `?pane=&tab=` to reset the pane and tab.

Closes #1076 
Closes #1078

### DIFF
--- a/src/components/account/SavedLocationsPage.js
+++ b/src/components/account/SavedLocationsPage.js
@@ -286,7 +286,7 @@ const LocationRow = ({ location, listId, typesAccess }) => {
       }}
     >
       <LocationLink
-        to={`/locations-standalone/${location.id}/${viewToString(location.lat, location.lng, MIN_LOCATION_ZOOM)}`}
+        to={`/locations/${location.id}/${viewToString(location.lat, location.lng, MIN_LOCATION_ZOOM)}?pane=standalone`}
         onClick={handleLocationClick}
       >
         <LocationTypeDisplay location={location} typesAccess={typesAccess} />
--- a/src/components/account/SavedLocationsPage.js
+++ b/src/components/account/SavedLocationsPage.js
@@ -40,7 +40,7 @@ const TopRow = styled.div`
   display: flex;
   align-items: center;
   gap: 8px;
-  padding: 0 1.25rem;
+  margin-block-end: 0.5em;
 `
 
 const ListName = styled.h2`
@@ -53,7 +53,6 @@ const ExpandRow = styled.div`
   display: flex;
   align-items: center;
   gap: 6px;
-  padding: 0.6rem 1.25rem;
   cursor: ${({ $clickable }) => ($clickable ? 'pointer' : 'default')};
 `
 
@@ -98,8 +97,8 @@ const LocationItem = styled.li`
   display: flex;
   align-items: center;
   justify-content: space-between;
-  padding: 0.5rem 1.25rem;
   border-bottom: 1px solid #efefef;
+  padding-block: 0.5em;
   position: relative;
 
   &:last-child {
@@ -111,7 +110,6 @@ const LocationLink = styled(Link)`
   color: ${theme.blue} !important;
   text-decoration: none;
   flex: 1;
-  font-size: 1rem;
 
   &:hover {
     text-decoration: underline;
--- a/src/components/activity/ActivityDiary.ts
+++ b/src/components/activity/ActivityDiary.ts
@@ -211,14 +211,11 @@ function transformActivityData(
           .map((typeId) => typesAccess.getType(typeId))
           .filter(Boolean)
 
+        const view = viewToString(change.lat, change.lng, MIN_LOCATION_ZOOM)
         const locationTypes: LocationTypes = {
           locationId: change.location_id,
           types,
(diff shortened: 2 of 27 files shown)


ezwelty:
@wbazant Thanks for going neck-deep to investigate!

#1070

* [x] App doesn't crash 🥳 
* [x] Changing the language again does not trigger a refresh, leaving the map and places in the previous language. Occurs in both mobile app (no restart) and web (no toast).

#1068

* [x] No (visible) URL redirect chain 🥳 
* [x] On desktop, it takes 4 clicks of browser back button to go back from location to `/changes`, `/users/:id/activity`, or `/account/lists`, and 2 clicks of back button if I try the same location a second time (?), while on mobile it is either 4 clicks if I access the list without visiting the map, and 1 click otherwise?
* [x] On mobile, if I then click on "go to map" from the location page, then click browser back button, it goes back to the original list, not the full location. And conversely, I can't get back to the original list if I reopen the drawer and click the back arrow (since now the arrow means close the drawer).

* [x] #1076
* [x] #1078

Some additional thoughts on the URLs:

* Frankly, I was always skeptical of the `@{lat},{lng},{zoom}z` syntax because it seemed like trouble to do something so custom. Without the @, couldn't it lead to issues with ambiguous routing in the future? Is there a good reason not to treat these as standard URL params? `?y={lat}&x={lng}&z={zoom}`? And I'm still in favor of dropping all the params into the browser bar (then no need for a share link button in the web version!).
* I notice that map `{lat},{lng},{zoom}z` and now also `pane` often end up passed along when navigating away from the map. Is this a bug or needed for a feature?
* If `pane=standalone` is what we use to link to locations, shouldn't that be the default? Then mobile `pane` can have three values `low`, `half` (or `mid`), `full`? Or would we rather link to the half drawer (the `pane` default) but can't because it has no back arrow? How about using standalone as full, and adding the standalone's back arrow to the half drawer in order to navigate back from it!? (the placemark button of standalone can be used to return to half drawer).

wbazant:
Thanks! I made fixes and follow-up issues.

> And I'm still in favor of dropping all the params into the browser bar (then no need for a share link button in the web version!).

That's the advantage, I've also had trouble keeping state in URL params because there are clashes on async updates. For e.g. map type, it would work, but then I'm not sure why e.g. the back button should bring you back to the previous setting.

ezwelty:
> @wbazant I'm not sure why e.g. the back button should bring you back to the previous setting

I think the back button should bring one back to the previous setting, as the result is a different map. Two map platforms I use all the time, https://map.geo.admin.ch/ and https://www.google.com/maps, behave this way.

Issue #1072, opened 2026-05-15
Keyboard covers new list creation


ezwelty:
The text box for entering a new list from the location page is covered by the keyboard that pops up to allow typing. It isn't in the page flow, and so the keyboard doesn't know to push things up. Fixes could include putting the popup content in a modal, putting the input for new list into the main location page content, and/or a multi <select> for (de)selecting existing lists (which would make list selection for touch friendly, but doesn't address the issue of creating a new list).

wbazant:
I could reproduce the bug in a location 2110414 without any description or reviews, with the unusual circumstance of the save button being visible in the partially open drawer. On web mobile I can just about click or see it in Firefox, and in the app it's very visible but hard to click. Then I see keyboard over the drawer, and the drawer is messed up from the screen height change, doing a weird thing when being open from over the keyboard.

The report button is also broken in the partially open drawer - clicking the description to focus dismisses the drawer.

Did you only see this in the partially open drawer? If that's the whole bug, my proposed fix is to add a call to fully open the drawer - with animation, so it's clear to the user what happens - on save or report.

ezwelty:
@wbazant Unfortunately not. What you describe I never observed because my phone is too short: I never can see those buttons when the drawer is only partly-open. The bug I describe happens only in the app (not in browser on same phone), with a long location (photo, description, etc), such that the buttons are towards to the bottom of the screen with a fully-open drawer. If I click to create a list, the keyboard opens, and for some reason the layout does not slide up (as it does in the browser version) to keep the input visible.

wbazant:
I've merged that one as a related fix and also added a bit of padding to the bottom of EntryOverview, so the buttons don't appear in the safe area. I can't reproduce your issue - my keyboard slides up - can you screencast it?

 I've also found two other issues for locations that are too long to fit in the screen.

ezwelty:
@wbazant Here is a video. Again only on my phone (Android 12) when running as an app.

https://github.com/user-attachments/assets/b6928c78-39eb-4fb6-a994-c180a4a9f341

PR #1084, opened 2026-05-19
Avoid text boxes in partially open drawer

wbazant:
Possibly a bit of a weird response of the app, but the action is also weird (it's quite hard for me to click that button on the edge). There's no issue with the review button because it opens in new page, so I unconfidently decided to not do the same there.

For #1072


ezwelty:
@wbazant I think you mean closes #1072?

wbazant:
Yes, thanks! fixed issue description.

PR #1085, opened 2026-05-19
Better photos

### LINKED ISSUE #1080: Compress less for better photo quality
In the app, the photos get resized, currently to JPEG quality 80 and max size 1200px. We could make the default value higher - Ethan suggests 2048 px.
Also, we could make it configurable, with Compression: standard, high quality, and full size, as options.

Reasons to compress are storage on our end (I think we resize anyway in addition to thumbnailing?) and upload speed. That second one varies depending on how good the user's internet is, for remote observations the photo upload occasionally took forever for me.

Suggested in https://github.com/falling-fruit/falling-fruit-web/pull/1075.

wbazant:
Closes #1080 

I've assembled an experimental setup of objects of varying colors in diffuse lighting, took two pictures, and uploaded: https://falling-fruit-web.pages.dev/locations/1989825/@55.8258624,-4.2610278,18z .

 I can tell the better one from the lighting on the window frame for example. Sizes are 162kb and 460kb, so still the same ballpark - I think the slow uploads I originally tried to fix were from pictures weighing a few megabytes. Agreeing with you @ezwelty  that the resizing was too aggressive, thanks for pointing it out!

### DIFF
--- a/src/constants/photo.js
+++ b/src/constants/photo.js
@@ -11,4 +11,4 @@ export const PHOTO_MAX_DIMENSION = 2048
  *
  * @constant {number}
  */
-export const JPEG_QUALITY = 0.85
+export const JPEG_QUALITY = 0.8
--- a/src/constants/photo.js
+++ b/src/constants/photo.js
@@ -4,11 +4,11 @@
  * Photos are downsized (if needed) while maintaining their aspect ratio.
  * @constant {number}
  */
-export const PHOTO_MAX_DIMENSION = 1200
+export const PHOTO_MAX_DIMENSION = 2048
 
 /**
  * JPEG compression quality for photos, between 0 and 1.
  *
  * @constant {number}
  */
-export const JPEG_QUALITY = 0.8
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant To match the legacy website, the new API produces three sizes: 100 px (thumb), 300 px (medium), 2048 px ("original"), which are uploaded to Amazon S3. So 2048 px would be a reasonable max size for the frontend. Did you feel that 0.8 → 0.85 was worth it?

https://github.com/falling-fruit/falling-fruit-api/blob/bbe8db0b87edec9e80e7eb3f750d23ccce953296/helpers.js#L450

The reason the medium often looks fuzzy in the drawer/sidebar is that these are often wider (average width for phones nowadays is ~400 px) and, if the image is vertical, it is 300 px tall so quite a bit less than 300 px wide.

wbazant:
> Did you feel that 0.8 → 0.85 was worth it?

I've not tried to compare on compression quality, just took a chatgpt recommendation. We can go with just the max width change!

PR #1094, opened 2026-05-22
Simplify panes

### LINKED ISSUE #1089: Simplify panes
Ethan:
>  I can't get back to the original list if I reopen the drawer and click the back arrow (since now the arrow means close the drawer).

> If pane=standalone is what we use to link to locations, shouldn't that be the default? Then mobile pane can have three values low, half (or mid), full? Or would we rather link to the half drawer (the pane default) but can't because it has no back arrow? How about using standalone as full, and adding the standalone's back arrow to the half drawer in order to navigate back from it!? (the placemark button of standalone can be used to return to half drawer).

Let's have three panes, where the map button is contextual - if the back button goes somewhere other than map, also render the map button. Main novelty is to allow going list -> full pane -> low pane via map button click -> full pane via drag up -> back to list.

Also, allow going back to list in the embed view because why not. Also, remove a notion of a drawer being 'disabled' that upon inspection isn't any different to being in the top position.

Let a link like /locations/1989707 still open the location in the drawer set to middle position- that's the most enticing view for the location, the viewer sees both overview and map context.

wbazant:
Closes #1089

### DIFF
--- a/src/components/account/SavedLocationsPage.js
+++ b/src/components/account/SavedLocationsPage.js
@@ -284,7 +284,7 @@ const LocationRow = ({ location, listId, typesAccess }) => {
       }}
     >
       <LocationLink
-        to={`/locations/${location.id}/${viewToString(location.lat, location.lng, MIN_LOCATION_ZOOM)}?pane=standalone`}
+        to={`/locations/${location.id}/${viewToString(location.lat, location.lng, MIN_LOCATION_ZOOM)}?pane=full`}
         onClick={handleLocationClick}
       >
         <LocationTypeDisplay location={location} typesAccess={typesAccess} />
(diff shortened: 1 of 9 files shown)


ezwelty:
@wbazant Looks promising, but some more tuning may be needed.

* [ ] The result doesn't quite match what you described as the solution. Maybe you meant to remove `?pane=full` from the location links?
* [ ] In mobile layout, if the list is loaded first before loading the map (so not if e.g. recent activity, your activity, or saved locations are loaded directly from URL), then one click of browser back button is sufficient. In other cases, including always on desktop layout, then 2–4 clicks are needed to go back. I suspect opening the location half-open (and thus loading the map) will affect (maybe fix?) this issue.

wbazant:
Thanks! I think I was explaining why I wanted /locations/:locationId to go to the half open pane. I made a change to not add ?pane=full on desktop, which doesn't actually remove the problem of extra entries in history. Since the severity is low I'll leave it as follow-up.

PR #1097, opened 2026-05-22
Sticky header

### LINKED ISSUE #1077: Issues with iOS back navigation and top bar
The top bar on iOS is often in conflict with content. It looks like a grey bar was inserted to accommodate it, but it was not fixed to the top of the viewport, so that on pages that are scrollable, the grey bar leaves to the top of the page and we're back to the top bar content overlapping page content.

Simultaneously on iOS, unlike Android, their is no device-level back button. So on long, scrolling pages, it is annoying to have to go all the way back up to navigate back with the "< Back" link.

I propose solving both of these problems by (1) making the grey bar position absolute (so that it stays in place when the rest of the page is scrolled) and (2) also make the "< Back" link sticky at the top of the page, with white (non-transparent) background so that the rest of the page can scroll below it.

Image
Image
Image

wbazant:
Closes #1077

### DIFF
--- a/src/components/ui/PageTemplate.js
+++ b/src/components/ui/PageTemplate.js
@@ -40,6 +40,13 @@ const SafeAreaInset = styled.div`
   background-color: ${({ theme }) => theme.secondaryBackground};
 `
 
+const StickyHeader = styled.div`
+  position: sticky;
+  top: 0;
+  z-index: 10;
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant A nice addition to our long about pages.

PR #1098, opened 2026-05-22
Add headers that say the same thing as the link to the page

### LINKED ISSUE #1096: Titles on info pages
Some pages, e.g. Recent activity, have a header, whereas others, e.g. the data or press page, don't. Maybe we should have the headers on all / most pages.

wbazant:
Closes #1096


ezwelty:
@wbazant Seems reasonable. Thanks!

PR #1099, opened 2026-05-22
Scroll into view

### LINKED ISSUE #1072: Keyboard covers new list creation

wbazant:
Closes #1072

For me, this commit makes a difference between the keyboard impinging onto the field about third of the way and being fully visible. @ezwelty I'm definitely guessing and looking for cheap fixes here but would you try this one?

Also, I notice that the soft keyboard appearance also moves the image and tabs up, which is not possible with a normal scroll. It's a feature rather than bug. I had a go at a re-engineer of the location drawer
thinking it could be quite good for the full pane to behave like a normal website but it was unfortunately hard, not working, and I abandoned it.

### DIFF
--- a/src/components/entry/overview/SaveToListButton.js
+++ b/src/components/entry/overview/SaveToListButton.js
@@ -169,6 +169,7 @@ const SaveToListButton = ({ containerRef }) => {
   const wrapperRef = useRef(null)
   const newListInputRef = useRef(null)
   const addingNewSkeletonRef = useRef(null)
+  const addNewRowRef = useRef(null)
 
   const { lists } = useSavedLists(locationId)
   const { drawerFullyOpen, fullyOpenPaneDrawer } = useLocationPane()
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant It'll fail if the screen is a bit shorter than mine, but it does work for me. Interesting the differences between the app running Android Webview (left) and the Chrome app on my phone (right):

Screenshot_20260525-180658
Screenshot_20260525-181032


I feel this may ultimately require a larger dedicated control (e.g. modal) to be robust and touch-friendly, but I think it's fine launching the app with this and then see what feedback comes in?

wbazant:
Thanks! I made an issue for insufficient space in the drawer. It's an idea to put 'add new list' in a modal, maybe more accessible but less slick (or at least I think it's slick, maybe people will complain and prove me wrong) - I think with the added list appearing just next to where the user was typing, it's easy to understand what happened without big design elements like toasts.

PR #1108, opened 2026-05-28
Stop header from overflowing on narrow desktops

### LINKED ISSUE #980: Readjust the narrow desktop layout so there's no overflow after opening the menu

wbazant:
Closes #980


ezwelty:
@wbazant I confirm the socials are hidden in time before squishing with even the widest (Greek) language.

PR #1110, opened 2026-06-03
Say why the map is stuck when it is

### LINKED ISSUE #1109: app: the map is stuck
It got into a state. Sadly a high severity bug of no known cause or repro into the stuck state. I have a recording - the share coords don't update, i guess the callback isnt triggered.

https://github.com/user-attachments/assets/df5dcf90-857d-4635-8fd3-e4e519a06e0a

Fixed after changing spots and logging out and then in, but changing language didn't work.

wbazant:
Closes #1109 
In the map callback, we have these actions:
```
  const throttledDispatches = throttle((newView) => {
    dispatch(updateLastMapView(newView))
    dispatch(fetchLocations())
    dispatch(fetchFilterCounts())
  }, 1000)
```
 `updateLastMapView(newView)` is sync, and the other ones are made from createAsyncThunk, with .rejected handlers toasting the errors. The whole thing is also a callback / setTimeout in the throttle so the error boundary doesn't cover it. So, if  `updateLastMapView(newView)` fails, then the other dispatchers don't happen, and it calls  `src/utils/persistentStore.js` , which code that can throw even though it shouldn't throw for us.

I'm blaming 'somewhere there' for the problem, two possible ones are JSON.stringinfy or setItem. I can sort of reproduce the bug, putting throwing code like `JSON.stringify({ n: 1n }) `  inside setLastMapViewThrottledSync results in a silent failure and call to fetch locations not happening:

The change puts a try/catch about syncing to persistent store, so if it goes wrong, it will say so, and proceed, so the severity is reduced just to an unimportant feature not working and a repeated warning.

### DIFF
--- a/src/components/map/MapPage.js
+++ b/src/components/map/MapPage.js
@@ -3,6 +3,7 @@ import { useCallback, useEffect, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { useDispatch, useSelector } from 'react-redux'
 import { useLocation } from 'react-router-dom'
+import { toast } from 'react-toastify'
 import styled from 'styled-components/macro'
 
 import { VISIBLE_CLUSTER_ZOOM_LIMIT } from '../../constants/map'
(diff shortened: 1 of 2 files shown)


wbazant:
Harmless change and has benefits in some theoretical situation so I'll merge it but not confident it closes the issue

PR #1112, opened 2026-06-06
Don't rely on currentZoom in zoom callbacks, so we can zoom in and out quickly

### LINKED ISSUE #1111: Zoom button shouldn't wait for map update
Press the zoom out button three times very rapidly. The map will only zoom out once if you do it quick enough, as if ignoring that it was given three commands.

Relatedly to #1109, the zoom button stops working when the view stops updating - this is because the callbacks use the value of googleMap.getZoom bound to the component, which is as of the last render of MapPage. Consider these log statements:
```
@@ -211,6 +212,7 @@ const MapPage = ({ isDesktop }) => {
   } = useSelector((state) => state.map)
 
   const currentZoom = googleMap?.getZoom()
+  console.log("MapPage", currentZoom)
 
   const place = useSelector((state) => state.place.selectedPlace)
 
@@ -380,7 +382,9 @@ const MapPage = ({ isDesktop }) => {
     googleMap?.setZoom(currentZoom + 1)
   }
   const zoomOut = () => {
+    console.log("zoom out", currentZoom)
     googleMap?.setZoom(currentZoom - 1)
+    console.log("zoom out after", googleMap.getZoom())
   }
 
   const toggleShare = useCallback(() => {
```

When the callback doesn't update the view, we see:
```
MapPage undefined
MapPage undefined 
MapPage 14
MapPage 14 
zoom out 14 
zoom out after 13
zoom out 14 
zoom out after 13
```

wbazant:
Closes #1111

### DIFF
--- a/src/components/map/MapPage.js
+++ b/src/components/map/MapPage.js
@@ -377,10 +377,14 @@ const MapPage = ({ isDesktop }) => {
   }, [googleMap, getGoogleMaps, isViewingLocation, history])
 
   const zoomIn = () => {
-    googleMap?.setZoom(currentZoom + 1)
+    if (googleMap) {
+      googleMap.setZoom(googleMap.getZoom() + 1)
+    }
   }
   const zoomOut = () => {
-    googleMap?.setZoom(currentZoom - 1)
+    if (googleMap) {
+      googleMap.setZoom(googleMap.getZoom() - 1)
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant Such a simple change, but it feels much snappier. Yeah!

PR #1113, opened 2026-06-06
Directly use idle handler

### LINKED ISSUE #1109: app: the map is stuck

wbazant:
Closes #1109


ezwelty:
@wbazant Thank you for digging into this and finding a likely solution. This feels like a good moment to release the app!

ezwelty:
@wbazant Merging in. Didn't see any issues during testing.

PR #1118, opened 2026-06-11
Delete account page

### LINKED ISSUE #1117: Add account deletion button
I should have seen this coming because it is an iOS App Store requirement. We need to add a button for a user to delete their own account. Easiest is probably adding a link to the list of links on the account page, followed by a confirmation dialog. The API already supports this with an authenticated `DELETE /user`.

# Notice from Apple:

### Issue Description

The app supports account creation but does not include an option to initiate account deletion. Apps that support account creation must also offer account deletion to give users more control of the data they've shared while using an app.

Follow these requirements when updating an app to support account deletion:

- Only offering to temporarily deactivate or disable an account is insufficient.
- If users need to visit a website to finish deleting their account, include a link directly to the website page where they can complete the process.
- Apps may include confirmation steps to prevent users from accidentally deleting their account. However, only apps in highly-regulated industries may require users to use customer service resources, such as making a phone call or sending an email, to complete account deletion.

### Next Steps

Update the app to support account deletion. Once account deletion has been added, or if it is already in place, reply to this message with a screen recording captured on a physical device that demonstrates:

- Creating a new account or signing in with the demo account
- Navigating to the account deletion option
- The complete account deletion flow from initiation to confirmation

Include the recording in the Notes field of the App Review Information section in App Store Connect for future submissions.

If the app is unable to offer account deletion or needs to provide additional customer service flows to facilitate and confirm account deletion, either because the app operates in a highly-regulated industry or for some other reason, reply to App Review in App Store Connect and provide additional information or documentation. For questions regarding legal obligations, check with legal counsel.

### Resources

Review frequently asked questions and learn more about the [account deletion requirements](https://developer.apple.com/support/offering-account-deletion-in-your-app).

wbazant:
Closes #1117


ezwelty:
@wbazant Thanks! Tested in desktop and mobile.

> This will permanently delete your account. Pages associated with your account will no longer exist. Any locations and reviews you contributed will become anonymous.

It wasn't immediately clear to me what "pages" referred to (location page?). How about something like:

> This will permanently delete your account and associated pages (profile, activity). Any locations and reviews you contributed will become anonymous.


Edit: Also, looking at this block of links with "delete" hiding in the middle, I wonder whether "Your activity" and "Saved locations" should move somewhere. "Your activity" could be under "Profile", but since "Saved locations" is private, maybe they could both go directly under "Account" ?

Screenshot 2026-06-15 at 22 53 07

wbazant:
I tried your verbiage but it translated weird and I propose to drop the sentence about pages, maybe if people have account deletion questions we will revisit it, but for now it's just for Apple to like:

'
This will permanently delete your account.

Any locations and reviews you contributed will become anonymous.
'


You're right about the links clashing, I had a few thoughts:
 - separating out the two groups would make the menu easier to understand
 -  the 'delete account' adds danger to all the neighbouring clicks too.
 -  the change password/change email / delete account are provided because they have to be there in case of a rare need, but the your activity and saved locations links go to useful features that might be used often
  
 I added a spacer between the two groups.

PR #1133, opened 2026-07-02
Hints

### LINKED ISSUE #1119: do something about old locations that improves the map over time
The project has been successful for a long time, with many locations imported in the early years, that are now over ten years old and sometimes not there any more. Users who join in that area can improve the freshness of data:

- report that they don't exist any more
- leave a review

It's easy to add quick edit options for these - open the forms prefilled.

We talked about it during the call. Some thoughts: "report" for locations that don't exist any more is not obvious, it conflates moderation with the locations going away over time. I got the a similar suggestion independently (to mark locations as stale and add a simple "yes, still exists", "no, does not exist" to them ).

Now I'm thinking, it's quite a heavyweight intervention, and it's going to be super annoying to see the alert, also discouraging to see everything is stale. Only someone who accepted the mission of improving freshness of the local map will want to have the quick edit options on.

So I'm thinking this is a quest / challenge someone can be invited to and then optionally take on. Like Falling Fruit Quests / Challenges that people get badges on except that's quite infantile. The other idea is regional content moderators who get given the power to delete / archive locations, except that adds a lot of complication - invite them by email, let them opt in, let them leave the role, train them to do the role right - and we still only have one Ethan's worth of moderation in the project.

Another idea is quick edit options which can be enabled in settings, would need #944 . That would be a bit of a nudge for people to give themselves quests, and then we can ask everyone in a newsletter to opt in if they want to.

wbazant:
Closes #1119

### DIFF
--- a/public/locales/ar.json
+++ b/public/locales/ar.json
@@ -182,12 +182,13 @@
     },
     "hints": {
       "add_description": "إضافة وصف",
+      "add_season_end": "إضافة نهاية الموسم",
       "add_season_start": "إضافة بداية الموسم",
-      "add_season_stop": "إضافة نهاية الموسم",
-      "edit_description_eg_still_there": "تعديل الوصف: 'لا يزال موجوداً'",
+      "confirm_still_exists": "تأكيد أنه لا يزال موجوداً",
       "over_x_years_ago": "منذ أكثر من {{yearsAgo}} سنة",
-      "report_comment_placeholder_no_longer_exists": "لم يعد موجوداً",
-      "report_no_longer_exists": "الإبلاغ: لم يعد موجوداً"
+      "report_comment_placeholder_no_longer_exists": "لم يعد موجودًا",
+      "report_no_longer_exists": "الإبلاغ عن عدم وجوده",
+      "review_comment_placeholder_still_exists": "لا يزال موجودًا!"
     },
     "index": {
       "editmarker_html": "<b>انتقل إلى موضع المصدر.</b><br>تحقق من عرض القمر الصناعي - قد يكون المصدر مرئيًا من الفضاء!"
@@ -220,7 +221,7 @@
       "added_by": "أضيف بواسطة {{name}}",
       "added_by_you": "أضيف بواسطتك",
       "date_added": "تمت الإضافة {{date}}",
-      "date_last_edited": "آخر تعديل {{date}}",
+      "date_last_updated": "آخر تحديث {{date}}",
       "imported_from": "تم الاستيراد من {{name}}",
       "season": {
         "in_season": "في الموسم من {{start_month}} إلى {{stop_month}}",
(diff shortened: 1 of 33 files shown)


ezwelty:
@wbazant Finally took it for a spin!

* [x] I encountered the "Add description" frequently, and clicking it works as expected.
* [x] Found "Add season stop" at 1919720 and works as expected.
  * [x] It is `season_stop` in the database, but a better wording would be "Add season end"
* [x] Found "Add season start" at 1554 and works as expected.
* [x] The season hints need some layout tweaks. Top-vertical align the calendar icon and align start of new line with start of first line?

Screenshot 2026-07-07 at 22 04 43

* [x] I think "Report: no longer exists" is better as "Report no longer exists"
* [x] I think "Edit description: still there" is better as "Confirm still exists" and link to the review form. I understand that you were avoiding the review form (which could lead to fragmented reviews), but reviews seem a better place for dated observations? Especially if we finally sort them by most recent first.
* [x] I was somewhat surprised not to see "Add photo". Again to avoid the fragmented review problem?

ezwelty:
@wbazant One more thing. Should we have something like a "Verify and edit" to unverified locations? Usually this is either the location is wrong or it has a single type that may be wrong, or it has multiple possible types, all but one is wrong. Unverified locations may also be candidates, regardless of age, for something "Report does not exist"

wbazant:
>  The season hints need some layout tweaks. Top-vertical align the calendar icon and align start of new line with start of first line?

That's better! It degrades a bit if you make the window extra narrow in https://issue-1119.falling-fruit-web.pages.dev/locations/1919720 but I'm not sure how to wrap the hint, but not the text afterwards.

>  better wording would be "Add season end"
>  I think "Report: no longer exists" is better as "Report no longer exists"

Thanks! Made changes and re-translated.

> I think "Edit description: still there" is better as "Confirm still exists" and link to the review form. I understand that you were avoiding the review form (which could lead to fragmented reviews), but reviews seem a better place for dated observations? Especially if we finally sort them by most recent first.

Changed this! Not sure what you mean about fragmented reviews, I was just trying to make the interaction make sense: the hint shows next to the date, so the action should make the date go up. The solution of changing the description achieved that but I agree 'still exists' is an observation. I switched to leaving a review, and introduced a 'last updated date' which is a max of last edit and reviews. Then after leaving the review, the date goes up, and the value is also more useful at assessing location freshness (it's now the last recorded interaction with it more generally).


>  I was somewhat surprised not to see "Add photo".

I ended up doing hints that fill in gaps and can't be turned off, this one would be quite annoying and not easy to bulk fix! Maybe one day as part of 'you approached this location' experience when we go fully pokemon go?

> Should we have something like a "Verify and edit" to unverified locations? 

Yeah would be good for explaining what 'unverified' means, too? I actually don't totally know myself what it means for imports. I made a demo so we can decide on the options shown, #1141

wbazant:
I cleaned up the code a bit. I'll merge this one, and we can work on more hints as we get clarity on what options to offer.

PR #1135, opened 2026-07-05
Panorama labels

### LINKED ISSUE #1081: Add labels to panorama (google street view)
For example, here is my local hill of apple and pear trees grown from pips under a campus:

https://beta.fallingfruit.org/locations/2112613/panorama/55.8252416,-4.2616985,18z

In that view, I would like to see labels. Also replacing the marker with our custom marker would be a nice touch.

Related to #647

Sveder:
closes https://github.com/falling-fruit/falling-fruit-web/issues/1081


Sveder:
Looked at the code, looks mostly fine to me. ptal :)

public/selected_location_pin.svg
wbazant, line 1: Instead of duplicating the code for an SVG can you do something so it references the imported SVG? I think src="... somehow definition of the svg here instead of the path... " is possible.

src/components/map/locationMarkerHelpers.js
```diff
@@ -0,0 +1,171 @@
+import { MapType } from '../../constants/settings'
+import { theme } from '../ui/GlobalStyle'
+
+// Shared marker + label rendering used by both the map (LocationMarkers) and
+// the Street View panorama (PanoramaHandler), so both look identical.
```
wbazant, line 5: Could you remove all LLM comments from your change, and all comments that explain what the code does? Comments that explain why the code does something can stay if hard-won, like if we needed to fix a bug caused by a gotcha.

src/components/map/locationMarkerHelpers.js
```diff
@@ -0,0 +1,171 @@
...
+const createTextShadow = (color, size = 1, times = 1) => {
+  const shadows = []
+  for (let i = 0; i < times; i++) {
+    shadows.push(`0px 0px ${size}px ${color}`)
+  }
+  return shadows.join(', ')
+}
+
+const getLabelStyleConfig = (mapType) => {
+  const configs = {
```
wbazant, line 38: Could this be an if statement?

src/components/map/locationMarkerHelpers.js
wbazant, line 1: Can you pick a name without "helpers" that corresponds to something on the site?

src/components/map/PanoramaHandler.js
```diff
@@ -3,16 +3,106 @@ import { useTranslation } from 'react-i18next'
...
+// The orange "here" pin drawn above the selected location's dot. Uses the same
+// boxicons Map glyph (in theme.orange) as the map's MapPin in Pins.js, anchored
+// at the tip so it floats above the dot.
+const getSelectedPinIcon = (googleMaps) => ({
+  url: '/selected_location_pin.svg',
+  scaledSize: new googleMaps.Size(48, 48),
+  anchor: new googleMaps.Point(24, 44),
+})
+
+// Dedupe the nearby locations by id (the viewed location may also appear in
```
wbazant, line 26: That's not right, the selected location will definitely be in the map locations list - see redux. You can just use the selected location to draw the pin and the other locations for dots.

src/components/map/PanoramaHandler.js
```diff
@@ -3,16 +3,106 @@ import { useTranslation } from 'react-i18next'
...
-  constructor(googleMap, googleMaps, location) {
+// Draws the same dot markers and labels as the map (LocationMarkers) onto the
+// Street View panorama, so both views look identical. The selected (viewed)
+// location additionally gets the orange "here" pin above its dot, mirroring
+// the map's SelectedLocation pin.
+class PanoramaWithMarkers {
+  constructor(googleMap, googleMaps, options) {
     this.googleMap = googleMap
     this.googleMaps = googleMaps
+    this.centerLocation = options.centerLocation
```
wbazant, line 48: Can you check how I call this concept elsewhere in the code? I think I went for 'selected location'.

src/components/map/PanoramaHandler.js
```diff
@@ -3,16 +3,106 @@ import { useTranslation } from 'react-i18next'
...
+
+      const labelData = (location.type_ids || [])
+        .map((id) => getDisplayLabel(this.typesAccess, id))
+        .filter(Boolean)
+      if (labelData.length > 0) {
+        const labelHtml = formatLabelHtml(labelData, this.selectedTypes)
+        // Always use the satellite/hybrid label style (white text, dark
+        // outline) since the panorama is imagery — more readable than the
+        // road-map style regardless of the map's current type.
+        const label = createLabel(
```
wbazant, line 94: Change the interface of createLabel so it accomodates this concept rather than passing MapType.Hybrid here.

src/components/map/PanoramaHandler.js
```diff
@@ -21,29 +111,33 @@ class PanoramaWithMarker {
...
 
     const panoClient = new this.googleMaps.StreetViewService()
     try {
       const panoData = await panoClient.getPanorama({
-        location: this.location,
+        location: this.centerLocation,
         radius: 50,
       })
 
+      // disconnect() may have run while we awaited getPanorama (a newer
```
wbazant, line 122: Not sure how would this happen from interactions in the UI

wbazant (changes requested):
Thanks! I had some comments to help the change fit better with the rest of the code base but it's otherwise all right and it looks like it will do the right thing. Once you addressed these requests I'll take it for a spin.

src/components/map/PanoramaHandler.js
wbazant: Actually since createLabel just has a big switch statement inside, we could have createWhiteTextLabel and createBlackTextLabel, call always the white one in PanoramaHandler, and have a conditional on what to call inside LocationMarkers

public/selected_location_pin.svg
wbazant: Sadly wasn't, left the comment in the SVG

wbazant:
Thanks! I thought I can do the cleanup for the markers code, I'll take it from here, see #1142.

PR #1141, opened 2026-07-09
Unverified with actions

### LINKED ISSUE #1140: More explainable 'unverified' with options
Ethan:

>  One more thing. Should we have something like a "Verify and edit" to unverified locations? Usually this is either the location is wrong or it has a single type that may be wrong, or it has multiple possible types, all but one is wrong. Unverified locations may also be candidates, regardless of age, for something "Report does not exist"

_Originally posted by @ezwelty in https://github.com/falling-fruit/falling-fruit-web/issues/1133#issuecomment-4912497062_

wbazant:
Closes #1140

Example: https://issue-1140.falling-fruit-web.pages.dev/locations/2111537/55.8275615,-4.2503360,18z

### DIFF
--- a/public/locales/ar.json
+++ b/public/locales/ar.json
@@ -194,7 +194,7 @@
       "unverified_edit_position": "تعديل الموقع",
       "unverified_edit_types": "تعديل الأنواع",
       "unverified_mark_as_verified": "وضع علامة كمتحقق",
-      "unverified_may_be_inaccurate": "لم يتم التحقق - قد يكون غير دقيق",
+      "unverified_may_be_inaccurate": "لم يتم التحقق: قد يكون غير دقيق",
       "unverified_report_does_not_exist": "الإبلاغ عن عدم وجوده"
     },
     "index": {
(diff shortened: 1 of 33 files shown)


ezwelty:
@wbazant Why did you close this? It was marked as a draft, so I was waiting to weigh in, but I think there's potential.

wbazant:
Sure i can ask for input more explicitly next time! The branch is toast but easy to reprompt, let's finish it here. I also liked making "unverified" actionable but wasn't hugely sure what it even means now. Do you have ideas for verbiage - menu header and then suggested actions?

ezwelty:
@wbazant I don't think there is much to change. The old website explained the unverified toggle as "If you doubt the existence, location, or identity of this source" (where "identity" means location types) and your actions cover these scenarios.

1. I don't have a strong opinion, but would consider somewhat simpler text for the existing options:
  * Could be inaccurate → May be inaccurate
  * All is good, verify → Mark as verified
  * Correct the position → Edit position
  * Correct the types → Edit types
  * Report does not exist (unchanged)
2. The "All is good, verify" option should link to the Unverified toggle in the edit-location form.

wbazant:
Thanks, I think we're aligned on this one, and I think the verbiage with actions is a better fit than the Unverified pill. For verifying, I made it happen just after confirm().

PR #1142, opened 2026-07-13
Panorama

### LINKED ISSUE #647: Add full access to Google Street View

### LINKED ISSUE #1081: Add labels to panorama (google street view)

wbazant:
Closes #1081
Closes #647.

Based on #1135 , credit to @sveder1 for starting this off!


ezwelty:
@wbazant The website and API are back!

How come you opened a new PR rather than commit to the existing PR/branch? In order to trigger a Cloudflare Pages preview?

wbazant:
Didn't hugely think about it, just had all these crappy commits and had to put the branch online somewhere! But seeing the preview is a bonus - currently the labels don't go over the other markers on hover as a regression.

More great bugs: panorama on desktop doesn't let you click on locations, and the labels somehow come off when you move around. Also, clicking 'save' reloads it back to where it was first opened.

wbazant:
This is now more ready! I still need to translate verbiage and check on an actual phone.

wbazant:
@ezwelty Have a look!

Features:
- remove Google Street View link from location pages
- add the pegman control to the map (if zoomed in to level where locations are shown, and not editing position on mobile)
- use the same SVG markers between map and panorama: dot for location, pins for positions of selected location
- move the 'update position' mechanics to google maps event listeners: update on drop
- disallow submitting location form if position changed and street view is open and show a warning that street view positions are inaccurate
- clicking location dot selects it, now also in street view
- set only full screen street view on mobile, disallow full screen on desktop
- clicking the address link on location page moves the street view
- if opening street view and a location is selected, try to face the location from wherever the pegman was dropped (on actual mobile, this only works if 'move phone to move panorama' is disabled)
- apply 'show labels' settings in street view, interpreting 'when zoomed in' as a yes for street view
- when moving around street view, fetch locations when coming to the edge of what's fetched. Show labels only within a select radius

ezwelty:
@sveder1 @wbazant Thanks for all this! I can see a lot of work went into this. Fun to play around with, especially on desktop where you have both street view and side pane in parallel and everything working together. Probably also fun in-situ on mobile surrounded by a tree inventory, but I'm not currently somewhere where I can test that.

1. Seems like Street View was a beast to work with. Did you ever consider using the default `map.getStreetView()`, which supports overlays and automatically shares markers (https://developers.google.com/maps/documentation/javascript/streetview)? I used that for the old (admin-only) street view and remember it mostly working. Maybe a bit late but mentioning in case it can make life easier.
2. I miss the Google Street View button on location page to quickly open Street View centered on the location (and quickly close Street View again). I think it would remain useful alongside the new controls.
3. Why not show pegman control on mobile during edit-position? Currently Street View cannot be used on mobile to fine-tune / preview position.
4. I like your implementation of them, but I have to wonder whether the position warning and blocker are too stringent. Street view georeferencing varies by region, but is probably never (?) worse than adding a location based on the low-detail Google roadmap. Which makes me wonder – should we show satellite basemap as default and drop the long "Check the satellite view..." text?
5. I think it would be fine to make the "move phone to move panorama" feature off by default. I found it annoyingly jittery and preferred pressing the button (above the zoom controls) to disable the feature.

#### Quirks

* [ ] Labels (from distant or dropped markers?) often end up in the top left corner of the window. For example here: https://issue-1081.falling-fruit-web.pages.dev/locations/3383/40.0405997,-105.2631823,19z (no way to link to a street view, correct?)
* [ ] Closing Street View (after potentially wandering a long way) returns to an unchanged map. This is perhaps desirable, but this confused me after selecting a location, closing Street View, and no longer seeing the marker for the selected location.
* [ ] Word wrap of labels as they approach the edge of the screen feels weird. Can it be disabled (to also match their map behavior)?
* [ ] Clicking a location in mobile layout does not do anything. Open the drawer?

wbazant:
Thanks for the comments, I just had a look again to see if there's something worth discussing at the meeting!

I'm adding edit position street view now, because indeed why not allow it. There's a clash with how we add the marker - moving the map under the centre is practical, but if it works differently between map and street view, it's easy to lose the position. Making the marker always be movable by drag is a tempting way out.

> I miss the Google Street View button on location page to quickly open Street View centered on the location (and quickly close Street View again). I think it would remain useful alongside the new controls.

I thought good riddance to it because it can't be made to work reliably and involves an occasional "we can't actually display it after all, disable", but you've convinced me over the past two weeks. How about making it the only way to open street view on desktop when not editing the position? The pegman annoys me on the upper left, it's too prominent, and also sometimes I try to drop the pegman near the location, and I can't see the location.

> Street view georeferencing varies by region, but is probably never (?) worse than adding a location based on the low-detail Google roadmap.

So the problem is that street views are inconsistent with themselves - see 2295626, strawberries under the fence, as you go around the corner. The dot is never where the fence is, but also shows different spots depending from where you look at it.

> Which makes me wonder – should we show satellite basemap as default and drop the long "Check the satellite view..." text?

I'm not sure - sometimes the satellite basemap is unavailable at high enough zooms, and it doesn't seem right to offer a toggle between basemaps that sometimes isn't respected, but I also see the practicality of the suggestion. 

I discovered that the banner helps me find the dot in street view. We can revise the verbiage - at least it shouldn't stay as is in street view because it makes no sense to send a street view user to take a look from space, but additionally, it could be different in edit position ("Move to the correct position"). Would it be good to shorten the banner everywhere?

wbazant:
For labels in top left corner, I have a repro - http://localhost:3000/map/55.8290222,-4.2614699,20z and drop to the roundabout, the first view faces north and you see apples that are in front, correctly, and also incorrect labels in the top left corner that correspond to locations behind, e.g. cherry. Rotating to the cherry and then back to the apples fixes it, or loading the street view before.

Also labels are often not keeping up with the rotations and other weird states happen if you do a lot to the map - 'quirks' is a great category for these. I'll merge just so there's an end but if you catch anything in another acceptance test let me know.

PR #1151, opened 2026-07-24
Redirect legacy /profiles/:id links to /users/:id

### LINKED ISSUE #1138: Missing redirect from `/profiles/id` to `/users/id`
Despite being short-lived, there are links to `/profiles/id` in the wild. We are missing a permanent redirect to `/users/id`.

CosmicLM:
Adds a redirect for the legacy /profiles/:userId path to the current /users/:userId profile route, following the same pattern as the other legacy redirects already in redirectRoutes.js 
(e.g. /imports -> /about/data).

Old /profiles/:id links still exist in the wild despite the path being renamed, and previously they'd hit a dead end instead of reaching the user's profile.

Closes #1138

### DIFF
--- a/src/components/init/redirectRoutes.js
+++ b/src/components/init/redirectRoutes.js
@@ -1,5 +1,5 @@
 import { useSelector } from 'react-redux'
-import { Redirect, Route } from 'react-router-dom'
+import { Redirect, Route, useParams } from 'react-router-dom'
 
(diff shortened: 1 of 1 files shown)


ezwelty:
@CosmicLM Thanks! Works and I added a regex to limit the redirect to integer :userId.
@wbazant Any possible downsides?

wbazant:
Not really, thanks guys!

PR #1152, opened 2026-07-24
Fix add similar issue for iOS

### LINKED ISSUE #1143: Add similar broken on iOS
After adding a location, 'add similar' prefills the form. This doesn't work on Michael's iPhone.

Can this be reproduced? I just checked on the webapp and it was fine.

https://github.com/user-attachments/assets/6ae7d7da-2cc8-4086-9906-b8086b440699

jordanstreiff:
Was able to reproduce the bug on my iPhone and simulator.

When user submits the location state was getting wiped because the effect was being rerun, which forces a refetch. If the data hasn't already been populated the blank state becomes the new state. So it's a race condition that could potentially effect more than iOS.

Fix tested / works in simulator.

Closes #1143

### DIFF
--- a/src/components/connect/ConnectNewLocation.js
+++ b/src/components/connect/ConnectNewLocation.js
@@ -36,7 +36,13 @@ const ConnectNewLocation = () => {
       // Should only happen for an artificially constructed URL
       history.push('/map')
     }
-  }, [dispatch, locationId, hasInitialView, isDesktop]) //eslint-disable-line
+    /*
+     * Mount-only: after a successful submit, locationId changes from 'new' to
+     * the new location's id while this component is still mounted. Re-running
+     * would call initNewLocation and wipe the just-added location from Redux,
+     * which the success page and 'Add similar' still need.
+     */
+  }, []) //eslint-disable-line
   return null
 }
 
--- a/src/redux/locationSlice.js
+++ b/src/redux/locationSlice.js
@@ -92,7 +92,15 @@ const locationSlice = createSlice({
     clearLocation: (state) => {
       Object.assign(state, initialState)
     },
-    duplicateIntoNewLocation: (state) => {
+    duplicateIntoNewLocation: (state, action) => {
+      /*
+       * The location captured when the success toast was created. Restore it
+       * in case something (e.g. the mobile drawer closing through /map)
+       * cleared the slice before the user tapped 'Add similar'.
+       */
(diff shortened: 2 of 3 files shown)


src/components/connect/ConnectNewLocation.js
```diff
@@ -36,7 +36,13 @@ const ConnectNewLocation = () => {
       // Should only happen for an artificially constructed URL
       history.push('/map')
     }
-  }, [dispatch, locationId, hasInitialView, isDesktop]) //eslint-disable-line
+    /*
+     * Mount-only: after a successful submit, locationId changes from 'new' to
```
wbazant, line 40: The comment helped me understand how the change worked, thanks! I'm trying to keep the codebase relatively comment free. Can we take it out before merging? Maybe see if the doc in connectRoutes.js needs revising.

src/redux/locationSlice.js
```diff
@@ -92,7 +92,15 @@ const locationSlice = createSlice({
     clearLocation: (state) => {
       Object.assign(state, initialState)
     },
-    duplicateIntoNewLocation: (state) => {
+    duplicateIntoNewLocation: (state, action) => {
+      /*
+       * The location captured when the success toast was created. Restore it
+       * in case something (e.g. the mobile drawer closing through /map)
+       * cleared the slice before the user tapped 'Add similar'.
+       */
```
wbazant, line 100: Would you check if changing the effect array in ConnectNewLocation is enough? This might not be necessary.

wbazant (commented):
Thanks for looking into this! Looks plausible, I think I also found a web repro - pull down the drawer and immediately press 'Add similar'. I'll build locally and check later when i have computer time. Happy to merge just after removing the explanatory comments but to maybe make this even better:
- Does changing the ConnectNewLocation effect array by itself fix the race condition on iOS?
- for the case when location ID changed to nothing (someone closed the drawer) and the toast is going away, maybe disable the button instead?

wbazant:
Superseded by #1166. My appetite for supporting the workflow of swiping the drawer and then pressing 'add similar' very quickly has gone away temporarily, so let's just modify the dependency array to [] because it only has effects that will happen once. Thanks again for figuring this out!

ezwelty:
Closing since we have #1166 but we can revisit if the new behavior doesn't feel satisfactory after continued use.

PR #1153, opened 2026-07-27
Check the env file is there just with Node

### LINKED ISSUE #1139: If missing the .env file, the app doesn't start and doesn't say why
I was just setting up on the new laptop, and forgot to set an `.env` file. The experience looked like `yarn start` taking forever to start and made me very confused. Probably also confusing to new contributors! Fix the start script.

wbazant:
Closes #1139

### DIFF
--- a/capacitor.config.ts
+++ b/capacitor.config.ts
@@ -2,7 +2,7 @@ import type { CapacitorConfig } from '@capacitor/cli'
 import * as dotenv from 'dotenv'
 
 // Load build options from .env file
-dotenv.config()
+dotenv.config({ quiet: true })
 
 const config: CapacitorConfig = {
   // android: uh.fallingfruit.app
--- a/craco.config.js
+++ b/craco.config.js
@@ -1,23 +1,25 @@
-class RequireEnvVarsPlugin {
-  constructor(requiredVars) {
-    this.requiredVars = requiredVars
-  }
+const dotenv = require('dotenv')
 
-  apply(compiler) {
-    compiler.hooks.beforeCompile.tap('RequireEnvVarsPlugin', () => {
-      const missingVars = this.requiredVars.filter(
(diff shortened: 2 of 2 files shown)


ezwelty:
@wbazant Wait, I had a thorough response to this, but apparently never posted it? I quickly recreate it...

Since we already check for required variables in `craco.config.js`, I would suggest simply refactoring that code so that the existing check runs inline:

```js
const dotenv = require('dotenv')

// Check .env file completeness
dotenv.config()

const REQUIRED_VARS = [
  'REACT_APP_API_URL',
  'REACT_APP_API_KEY',
  'REACT_APP_RECAPTCHA_SITE_KEY',
  'REACT_APP_GOOGLE_MAPS_API_KEY',
]

const missingVars = REQUIRED_VARS.filter(
  (varName) => process.env[varName] === undefined
)

if (missingVars.length > 0) {
  throw new Error(
    `Build failed: Missing required environment variables:\n${missingVars
      .map((v) => `  - ${v}`)
      .join('\n')}`,
  )
}

module.exports = {
  webpack: {
    configure: (webpackConfig) => {
      webpackConfig.ignoreWarnings = [
        function ignoreSourcemapsloaderWarnings(warning) {
          return (
            warning.module &&
            warning.module.resource &&
            warning.module.resource.includes('node_modules') &&
            warning.details &&
            warning.details.includes('source-map-loader')
          )
        },
      ]

      return webpackConfig
    },
  },
  eslint: {
    enable: true,
    configure: (eslintConfig) => {
      // Remove CRA's react plugin references
      delete eslintConfig.plugins
      delete eslintConfig.extends

      // Let your own .eslintrc.json fully define the config
      return eslintConfig
    },
    mode: 'file', // force ESLint to use your .eslintrc.json
  },
}
```

wbazant:
Thanks, that's much cleaner!

PR #1159, opened 2026-08-20
Add Wojtek as Technical Director on about page

ezwelty:
Also moved "Executive Director" to translations, removed board president role, and added "England" to advisor locations for consistency.

@wbazant Can you take a look, in particular at your bio (and translations thereof) and the crop and choice of photo?


wbazant:
Thanks! The AI translations seem fine, they're very literal but functional. I like being Dyrektor Techniczny and how that's not a thing that anyone has ever heard of but it carries the gravitas.

PR #1163, opened 2026-08-25
Replace static central pin with draggable markers on mobile

### LINKED ISSUE #1162: Edit position broken on mobile
Try edit position twice. The orange pin does not come back to where it should - are we using position from redux to render it?

Broken by https://github.com/falling-fruit/falling-fruit-web/pull/1142 .

Additionally:
 - I've left Add position with the central pin and logic to sync central position and marker, which is probably suspect
 - the tooltip has too much content for mobile

The proposed resolution is I think to standardise on the 'move to the position of the source' also for new locations, and make the tooltip shorter (keep the hint about satellite view on desktop). Check:

- open form, edit position between street view and marker
- open form, edit position, cancel, edit position again
- open form, edit position, confirm, edit position again
- load an edit position URL
- new location and edit location

wbazant:
Closes #1162


wbazant:
@ezwelty I've merged this to get to a clean branch, although the branch is admittedly broken because I've also earlier merged the google street view branch, started to work on something else, and found the problems. Do have a look anyway! I think the solution has downsides (dragging the marker is a harder gesture on mobile, maybe I'm exaggerating) but keeping the current one and having a different mechanism between map and street view is super confusing.

PR #1165, opened 2026-08-25
Skip degenerate bounds on map view change to prevent unnecessary refetch

### LINKED ISSUE #1164: On mobile, going map -> settings -> map needlessly flickers the locations
I think this got introduced after we switched to native markers.

wbazant:
Closes #1164

### DIFF
--- a/src/components/map/MapPage.js
+++ b/src/components/map/MapPage.js
@@ -150,7 +150,6 @@ const makeHandleViewChange = (dispatch, googleMap, history) => {
     const center = googleMap.getCenter()
     const bounds = googleMap.getBounds().toJSON()
 
-    // Skip degenerate bounds (e.g. map resized to ~1px when switching tabs on mobile)
     if (isDegenerate(bounds)) {
       return
     }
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant Tested and the flashing is gone!

But why is `GET /locations/api/0.3` called in the first place if nothing has changed (also when switching between map and list)?

wbazant:
Thanks! I think the bounds of the map change when you move to settings which is why the callback gets triggered - a bit like on desktop, we're making a window bigger. For list, ?photo=true loads the thumbnails.

ezwelty:
@wbazant I think it is deeper than that:

* Clicking a location calls `/locations` (and `/locations/counts` on desktop)
* Unclicking a location calls `/locations` (and `/locations/counts` on desktop)
* Opening Settings calls `/locations` (and `/locations/counts` on desktop)
* Going back from settings calls `/locations` (and `/locations/counts` on desktop)
* Going to list and back to map calls `/locations`

And this is without touching the map. The API calls are identical. Some of these seem understandable and harder to prevent without introducing a bug (e.g. back from list), but others I find somewhat surprising (e.g. opening settings in desktop or in mobile). Not a huge priority, but still a performance hit and maybe reveals an underlying logics issue?

PR #1168, opened 2026-08-25
Ignore marker clicks while editing location

### LINKED ISSUE #1167: Mobile: accidental marker tap navigates away while editing location
When editing a location on mobile, it's easy to accidentally tap a nearby marker and get navigated away, losing the edit session. I've also once forgot I was editing position, and the map has let me go away which was very confusing.

Simplest to just not react to the clicks when editing or adding, assuming they're usually accidental. There's also a confirm() option but probably still confusing to see a confirm() in response to an accidental click?

wbazant:
Closes #1167

### DIFF
--- a/src/components/map/MapPage.js
+++ b/src/components/map/MapPage.js
@@ -366,12 +366,22 @@ const MapPage = ({ isDesktop }) => {
 
   const handleLocationClick = useCallback(
     (location) => {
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant Thanks for spotting. Tested and found that it fixes the issue.

PR #1175, opened 2026-08-27
Harmonise translations of key terms

### LINKED ISSUE #786: Use consistent wording in French translation
The machine translations are pretty decent, but based on the French translation, they don't necessarily propagate patterns present in the existing translations. I notice that the same word in English is thus translated differently, which is a bit confusing.

Here is the beginning of a mapping that I could use when revising the translations:

* `location`: `site` (shorter than `endroit` or `emplacement`)
* `position`: `position`
* `review` (noun): `avis`
* `review` (verb): `évaluer`

wbazant:
Closes #786


ezwelty:
@wbazant Nice! Well at least in French I can confirm that it is more consistent, and the choice of terms is reasonable. But for this to stick, shouldn't we maintain a mapping from English, perhaps in the translation utility, that remembers for future translations? Something like:

```json
{
  "fr": {
    "location": "site"
   }
}
```

Otherwise looks good (to the best of my language abilities).

Edit: Or does your new translation prompt essentially ensure that new translations are consistent with existing translations for a given locale, and that translations for new locales are consistent when first created?

wbazant:
Thanks for taking a look! I think I'll mostly now use an agent to add translations, I added an entry point into translation manager that lets it prepare new translations as an additional new file, and sometimes the agent edits the files with sed type text manipulation which is silly to watch but works fine. To get verbiage, it seems to check the files for how things are already called without me instructing explicitly.

PR #1180, opened 2026-08-29
Align the unverified hint with overview content

### LINKED ISSUE #1178: Unverified text is misaligned
Using `/locations/2111537` as an example:
Image

The "Unverified" text is indented, while it should be aligned with the other text. 

Introduced by #1140.

lKolabrodl:
## Summary

- Remove the inline-only start margin from the standalone Unverified hint.
- Preserve the existing margin for the stale-location hint, which still renders beside the last-edited text.

Closes #1178

## Problem

`UnverifiedHintToggle` reused `OldLabelInline`. That shared component has `margin-inline-start: 0.5em` because the stale-location hint appears inline after a date. The Unverified hint is rendered on its own row, so the inherited margin indented it relative to the rest of the overview.

## Fix

Add a small `UnverifiedLabel` styled variant that resets only `margin-inline-start`, then use it for `UnverifiedHintToggle`. Logical margin properties are retained, so the alignment also works in RTL locales.

## Validation

- Reproduced on the #1141 preview for `/locations/2111537`: hint left edge `18.67px`, parent left edge `12px`, computed margin `6.67px`.
- Verified the reset produces matching `12px` left edges and `0px` computed margin.
- `yarn lint:check`
- `yarn knip`
- `npx prettier src/components/entry/overview/Hints.js --check`
- `yarn build`

### DIFF
--- a/src/components/entry/overview/Hints.js
+++ b/src/components/entry/overview/Hints.js
@@ -35,6 +35,10 @@ const OldLabelInline = styled(ResetButton)`
   }
 `
 
+const UnverifiedLabel = styled(OldLabelInline)`
+  margin-inline-start: 0;
+`
+
 const ExistenceActionsContainer = styled.div`
   display: flex;
   flex-direction: row;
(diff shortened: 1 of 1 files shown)


wbazant:
I checked out this branch locally and your change fixes the issue. Thanks for contributing!

> Reproduced on the https://github.com/falling-fruit/falling-fruit-web/pull/1141 preview for /locations/2111537: hint left edge 18.67px, parent left edge 12px, computed margin 6.67px.

Which agent do you use? Is this from a plugin that can compile the app, run it through the browser, and examine the rendered content? I've not tried it with Kiro mostly because of still being new to it but it would be super interesting.

PR #1182, opened 2026-09-01
Add invisible recaptcha for signed out users when marking as verified

### LINKED ISSUE #1179: Mark as verified fails when signed out
Results in error:

`Location editing failed: reCAPTCHA invalid-input-response`

Introduced by #1140.

wbazant:
Closes #1179


ezwelty:
@wbazant Thanks, tested (on web only, desktop and mobile) and works.

PR #1183, opened 2026-09-01
Hide init location nav in street view

### LINKED ISSUE #1177: Cannot exit street view when setting location position
The street view frame extends below the header, such that the back button is hidden beneath it. The best solution is probably to set the street view frame to stop where the header begins?

Image

Result of #1142.

wbazant:
Closes #1177


ezwelty:
@wbazant Thanks, that does the trick!

PR #1190, opened 2026-09-01
Order reviews latest-first

### LINKED ISSUE #1062: Location's representative photo
Right now the photo chosen is the earliest one, so for example a picture of a tree when first planted.

Make it instead the first photo of the latest review. Make the change in list view, and in the carousel.

### Discussed in https://github.com/orgs/falling-fruit/discussions/1020

<div type='discussions-op-text'>

<sup>Originally posted by **wbazant** March 27, 2026</sup>
Right now a picture of a tree when planted is always the first one. Surely the newest one makes more sense?</div>

wbazant:
Closes #1062

@ezwelty I think we want it but have a look! I just checked that it got implemented, the order of reviews is probably quite logical because the most interesting ones are seen first.

### DIFF
--- a/src/redux/locationSlice.js
+++ b/src/redux/locationSlice.js
@@ -42,6 +42,12 @@ const computeLastUpdatedDate = (locationData, reviews) =>
       locationData.updated_at,
     )
 
+const reviewDate = (review) =>
+  new Date(review.observed_on || review.created_at).getTime()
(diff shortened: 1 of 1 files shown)


ezwelty:
@wbazant I think this is long overdue, can't think of a downside. Thanks!

PR #1192, opened 2026-09-02
Make the lightbox full screen

### LINKED ISSUE #733: Improve mobile photo viewer: let modal fill entire screen and allow zooming into photos

wbazant:
Closes #733


ezwelty:
@wbazant Pretty slick! Works great for the photos.

* I didn't immediately grasp the meaning of the "fit to screen" icon, and I wonder if it is needed at all. Clicking on the photo (or zooming in) switches to full size, while zoom out returns to fit-to-screen. I would suggest making it so that clicking on the photo again (with zoom out cursor on desktop) fits to screen, and dropping the "fit to screen" button.
* One thing I do miss from the old lightbox is seeing the review alongside the photos (to know when they were taken, what was said about them, what fruiting status they represent, etc). This because useful with locations with multiple reviews with photos. Could the review still be rendered somewhere (bottom left or right depending on direction)? Or is that too clumsy?

wbazant:
Thanks, I thought it came out great too, clearly the agent trained on some fantastic code from somewhere. I've dispensed with the 'fit to screen' and added click out to zoom out, with a heuristic to distinguish clicks from drags.

I've also standardised the square icon buttons between here and close street view, and made them have the same offsets as round icon buttons. Also, I've added ease-in of images - show the medium size for a bit while loading the original.

Showing the connection with review is hard on mobile, I guess there is space for it on desktop but in the current UI it's probably just confusing when the user got to the lightbox from the carousel. If they got to it from the review, they remember what they just clicked and there's a 'close', so they know they can go back to that view.

Issue #787, opened 2025-05-02
Formalize and implement taxon naming conventions


ezwelty:
Still under construction...

## Checklist

_Resulting from the conventions outlined below._

### Database or API

- [ ] As hybrid symbol, replace letter (x) with the multiplication sign (×)
- [ ] Remove space after × if not separating two parents
- [ ] Replace _Malus pumila_ with _Malus domestica_
- [ ] Search for and resolve potentially problematic common names (for English, see https://www.publicgardens.org/wp-content/uploads/2023/10/potentially-problematic-common-names-north-american-public-gardens.pdf)
- [ ] Enforce hyphenation convention as described below
- [ ] Enforce capitalisation convention as described below

### Website

- [x] Display scientific name with italic font
- [ ] Display cultivar name with plain font (#762)
- [ ] Display hybrid symbol with plain font

### Format

* Higher taxa (italic, capitalised)
  * _Arecaceae_ (family)
* Genus (italic, capitalised)
  * _Prunus_ (genus)
* Subgenus (italic, capitalised after _subg._)
  * _Prunus subg. Amygdalus_
* Species (italic, lowercase)
  * _Prunus domestica_
  * _Ribes uva-crispa_
* Subspecies (italic, lowercase after _subsp._: subspecies, _var._: variety, _f._: form)
  * _Prunus domestica subsp. domestica_
  * _Prunus persica var. nucipersica_
  * _Brassica oleracea var. capitata f. rubra_
* Cultivar (plain, title case, in single quotes)
  * _Cornus kousa_ 'Wolf Eye'
  * _Malus domestica_ 'Cox's Orange'
  * _Prunus domestica subsp. syriaca_ 'Mirabelle de Nancy'
* Hybrid (denoted by plain multiplication sign ×, followed by space only when separating two parents)
  * _Prunus_ ×<i>eminens</i>
  * ×<i>Sorbopyrus irregularis</i>
  * _Prunus cerasus_ × _fruticosa_ (short for _Prunus cerasus_ × _Prunus fruticosa_)
  * _Malus_ × 'Dolgo'

This list omits at least a few potential components, which are currently unused for lack of need, practicality, or clarity.

* Authority name – after genus, subgenus, or species (plain, historical names abbreviated following convention)
  * _Prunus domestica_ L.
  * _Euphorbia pulcherrima_ Willd. ex Klotzsch
[...]

Issue #542, opened 2024-10-26
Make the list view less repetitive


wbazant:
When testing #525 in Boulder, CO (lots of similar types because it's a map import), the user was confused about the "repeated" locations in the list. One idea we had on the spot was to indicate direction as well as distance - e.g. ✵ 50 meters northwest - which we would internationalize, see #531.

On more reflection, since we have access to all the location data, we could also show something else with secondary text, like an address, or only show the directions when there's an obvious center to give these locations from: only when there's a searched place or when geolocation is on.

ezwelty:
@wbazant Interesting observation. My first instinct was to mimic the mobile app, which includes the beginning of the location description, with an ellipsis (...) if it doesn't fit on one line. Which could be a nice addition. But this doesn't really help in this case where the location types and descriptions are all the same.

If we were to mimic the mobile app's behavior of a small discrete '+' in the center of the map, then their would always be a reference for distance and cardinal direction. 🤷🏼‍♂️  Whatever the reference, it should probably be the same as the distance, which currently is always the current map center, regardless of place search or geolocation.

wbazant:
Also suggested in #416:
- Make photo thumbnails larger
- Make photo thumbnail placeholder less prominent

PR #1196, opened 2026-09-03
Refactor type code and show cultivars without cursive

### LINKED ISSUE #762: Follow botanical convention when displaying names of cultivars
Sometimes types have a scientific name and cultivar. We often render them like this:

**Apple 'Cox'** *Malus pumila 'Cox'*

It's wrong, should be **Apple 'Cox'** *Malus pumila* 'Cox' . See https://en.wikipedia.org/wiki/Cultivar#Presenting_in_text for convention.

There's a number of places - everywhere the code uses "scientificName" for display needs reviewing:
- in the type tree
- in the list
- in location drawer
- in type select dropdown
- on activity page

_Originally posted by @ezwelty in https://github.com/falling-fruit/falling-fruit-web/issues/749#issuecomment-2813279241_
        
Solving some of the issue with a partial PR is welcome!

wbazant:
Closes #762

### DIFF
--- a/src/components/account/SavedLocationsPage.js
+++ b/src/components/account/SavedLocationsPage.js
@@ -191,11 +191,11 @@ const LocationTypeDisplay = ({ location, typesAccess }) => {
   }
 
   const typeElements = typeIds.map((typeId, idx) => {
-    const label = typesAccess.getDisplayLabel(typeId)
-    if (!label) {
+    const type = typesAccess.getType(typeId)
+    if (!type?.displayLabel()) {
       return <span key={idx}>{typeId}</span>
     }
-    return <CommonOrScientificName key={idx} label={label} />
+    return <CommonOrScientificName key={idx} type={type} />
   })
 
   return (
--- a/src/components/account/SavedLocationsPage.js
+++ b/src/components/account/SavedLocationsPage.js
@@ -30,6 +30,7 @@ import { theme } from '../ui/GlobalStyle'
 import Input from '../ui/Input'
 import { Page } from '../ui/PageTemplate'
 import ResetButton from '../ui/ResetButton'
+import { CommonOrScientificName } from '../ui/TypeName'
 import withRedirectToAuth from './withRedirectToAuth'
 
 const ListCard = styled.div`
@@ -125,7 +126,6 @@ const LocationLinkSkeleton = styled.a`
 const Address = styled.span`
   color: ${theme.secondaryText};
   font-size: 0.85rem;
-  margin-left: 0.5rem;
 `
 
 const RemoveButton = styled(ResetButton)`
@@ -136,19 +136,11 @@ const RemoveButton = styled(ResetButton)`
   pointer-events: ${({ disabled }) => (disabled ? 'none' : 'auto')};
   transition: color 0.15s ease;
   flex-shrink: 0;
-  margin-left: 0.5rem;
+  margin-inline-start: 0.5rem;
 
   &:hover {
     color: ${darken(0.4, theme.secondaryText)};
   }
-`
-
-const ScientificName = styled.span`
-  font-style: italic;
-`
-
-const CommonName = styled.span`
-  font-weight: bold;
 `
 
 const SkeletonListCard = () => (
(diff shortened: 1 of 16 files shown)


wbazant:
@ezwelty So this just unitalicises the cultivar part. In the most common case where both common and scientific name are present, this does either
**European plum** *Prunus domestica* 'Stanley' (location page or types dropdown)
or
**European plum 'Stanley'** (map labels, activity pages, saved locations page).

Shall we make the full form coherent with the short form and stick the cultivar to the common name - 
**European plum 'Stanley'** *Prunus domestica*  ? 

Edit: I think to fully flesh out the proposed rule: the cultivar will try to follow common name. So it's there in bold if all are present, but for no common name, *Prunus domestica* 'Stanley' is still fine.

It might also be interesting to reserve the bold font for common names - items outside taxonomy like Dumpster could get a non-bold non-italic styling.

ezwelty:
> Shall we make the full form coherent with the short form and stick the cultivar to the common name - **European plum 'Stanley'** _Prunus domestica_ ? [...] for no common name, _Prunus domestica_ 'Stanley' is still fine.

Tacking the cultivar to the common name is accepted practice. The RHS does this: https://www.rhs.org.uk/plants/132217/prunus-domestica-stanley-c-d/details
It isn't strictly correct for the cultivar to be missing from the scientific name, but repeating the cultivar feels unnecessary. If we think it serves our users better, then I'm open to it. Would it still show in results for the search "Prunus domestica 'Stanley'"?

> It might also be interesting to reserve the bold font for common names - items outside taxonomy like Dumpster could get a non-bold non-italic styling.

Not sure about this one. I would argue "dumpster" is the common name for a non-living thing that thus lacks a scientific name. So we would rather do it to place emphasis on living things, but if so, then shouldn't scientific-only taxa (without a common name) be both italic and bold to remain emphasized?

wbazant:
Now with some behaviour changes:
- bold cultivar in tree select
- only show one name in list locations, giving it the same labels as the map
- in location overview, apply the logic we have for making good display labels to the common name line, so it shows up with cultivars

wbazant:
Happy to repeat the cultivar where both common and scientific name show, so both are complete names. I just realised I forgot to do it on the type search menu.

ezwelty:
Tried out the branch a bit. Thanks for the thoughtful refactoring.

* [x] Yay for name and font style conventions!
* [x] I like the addition of the cultivar to the common name for compact views (currently: map labels, activity feeds).
* [ ] In full view, I definitely wouldn't want the cultivar name to be repeated twice (once in common name, once in scientific). I'm open to placing the cultivar with the common name. If so, that would mean dropping cultivar from scientific name in location overview, and moving it from scientific to common name in the type dropdown. But I would only do this if we suspect this is better for most users (visual consistency, knowledge of cultivar names greater than scientific name) – otherwise, I would keep the cultivar with the scientific name. A compromise, which is growing on me but may be visually confusing in practice, would be to keep the cultivar with the scientific name in full view, but bold the cultivar name.
* [ ] Is it possible to bold the cultivar name but not the single quotes? I think this would cut down on "visual noise". But small potatoes.
* [ ] Can you please restore scientific names to the list view? If the list shows the same as labels, why have the list at all?

Regarding italics, note that the hybrid x (which should in the future come from the API as the more correct ×) should not be italicized either. See #787. Not sure how painful that is to implement (or a potential performance hit?).

wbazant:
Sure, we can skip the cultivar from scientific name if it's already with the common name, and then it also works for the list - for the list I thought it's definitely too repetitive so I took them out, but now they're back in.

 I think the fix for making the list good is more content #542, I'll get onto the issues which need both api and UI changes one day.

> only do this if we suspect this is better

I think it is, on UX grounds. Consider a user who has a goal, decides on an action, and gets feedback from the world, in the scenario of clicking a dot on the map and seeing the location open. They've just clicked on a location which was a dot with a bold display label, then a drawer opened, with another piece of bold text. When the two don't match it's like a branch misprediction in the user's cogs, they have to work out what the text means and why it's different.

For #787, I also implemented the correct rendering of hybrids - not very expensive to determine or show, and also wasn't a pain to implement, I just gave the agent your examples.

PR #123, opened 2021-05-21
🦾 Font scaling and other touch-ups

SirajChokshi:
## Status:

:rocket: Ready

## Description
- Moved all fonts to use `rem`
- Scaled down fonts on desktop
- Added hover effects to desktop Buttons
- Added hover effects to desktop ListEntries
- Replaced text-shadow with outline on Locations
- Corrected Textarea font-size and padding
- Resized ListEntry contents so lists appear less dense

## Todo
- Type Filtering Fixes:
  - [x] Fix padding between filter checkboxes and text
  - [x] Decrease text and checkbox size
- WebKit
  - [x] Resolve Missing Icons
  - [ ] Vertically center type-filter checkboxes
- [x] Add mask component for list loading
- [x] Change opacity of locations on the satellite layer
- [x] Correct irregular sizes

Fixes #91 and #96

PR #1203, opened 2026-09-11
Add titles to pages

### LINKED ISSUE #113: Update title of app according to page
maybe:
About | Falling Fruit
1991282 Blackberry | Falling Fruit

Ethan pointed out that titles of pages go into link previews in #636. Check how e.g. location pages look when sent via whatsapp!

- Use react-helmet

wbazant:
Closes #113


ezwelty:
@wbazant Looks good, I tried to break it and couldn't.

If we want the titles to be distinctive, I do wonder about the original suggestion of including the location ID in the title. We have that for imports (Import #{id}: {name}) but not for users or locations. At least for locations, omitting the id results in many pages with the same title.

wbazant:
Thanks! Added (Location #123) in brackets so it starts with the type name.

Issue #1204, opened 2026-09-11
Let app show locations without loading types first


wbazant:
We currently wait for 5000 types to load at the start, to then query /locations and /clusters. The failure to load types is fatal. This design will stop working one day, and adds a bit of slowness to the first load - it's probably something to at least start performance tracking.

We could ease the pressure by adding an endpoint to fetch types in a given language, or support queries with a default type value.

ezwelty:
Possible strategies:

* Add `ids=0|1|...` to `GET /types`
* Add `locales=fr|en|...` to `GET /types`
* Add `embed=types` to `GET /locations/{id}`

PR #1206, opened 2026-09-11
Improve load sequence

### LINKED ISSUE #880: Disable TrackLocationButton, zoom, and  map movements until map finishes its on load callback
Same scenario as #878 and #879 - initial load seems to take longer, user presses buttons, and digs themselves a hole.

Navigating to a location sort of works before the first map callback completes, by moving the dot, but not fully: the URL wasn't updating for me when I interacted with it on mobile and stuff was just half-working.

Better to avoid the half-working experience and limit the bug surface -  pressing the buttons causes callbacks that will normally happen after the first callback.

Buttons for zooming in, zooming out, and tracking location  are easy to disable: change opacity and set pointer-events to none. I think the google maps API will also offer disabling movements that we can add as a react prop.

We can use lastMapView being null as a criterion for callback having fired at least once - see src/components/map/MapPage.js .

wbazant:
Closes #880


ezwelty:
@wbazant Removing google-map-react seems like a big change, but I can't detect anything awry with the map. Any specific stress tests I should try? I can't replicate the original issue so hard to gauge the behavior change.

wbazant:
Thanks! I tried the things I could think of like resizing the window, switching on and off on desktop which removes and reinstatesthe map, or on mobile which hides it. I also had some Playwright tests generated which made things slower or broken, which didn't really find anything except give me the idea for #1204.

PR #1208, opened 2026-09-12
Upgrade react-split-pane from v2 to v3

### LINKED ISSUE #1207: Update react-split-pane
We're currently on v2. Looks like there's v3, and compatible with React 17: https://www.npmjs.com/package/react-split-pane

Let's switch to it, the current old version says it's incompatible with React 17 and throws off attempts to npm install.

nopaalh:
Closes #1207

## Summary
Upgrade `react-split-pane` from v2 to v3 to fix compatibility with React 17.

## Changes
- Bump `react-split-pane` to v3 in `package.json`
- Migrate `DesktopLayout.js` to the new v3 API:
  - Named imports: `{ SplitPane, Pane }`
  - `split="vertical"` → `direction="horizontal"`
  - Wrap children in `<Pane>`
  - Move `minSize` / `maxSize` / `defaultSize` to `<Pane>`
- Add Webpack alias for `react/jsx-runtime` in `craco.config.js` to fix ESM resolution under React 17

## Notes
- Unrelated warnings (Redux Toolkit deprecation, Google Maps billing) are out of scope.
- The `aria-description` warning originates from `react-split-pane` v3 itself and is non-fatal.

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -35,7 +35,13 @@ module.exports = {
             warning.details.includes('source-map-loader')
           )
         },
-      ]
+	]
+
+      webpackConfig.resolve.alias = {
+        ...webpackConfig.resolve.alias,
+        'react/jsx-runtime': require.resolve('react/jsx-runtime.js'),
+        'react/jsx-dev-runtime': require.resolve('react/jsx-dev-runtime.js'),
+      }
 
       return webpackConfig
     },
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "^5.0.1",
     "react-select": "^4.3.0",
-    "react-split-pane": "^0.1.92",
+    "react-split-pane": "^3.2.0",
     "react-toastify": "^8.1.0",
     "reaptcha": "^1.7.3",
     "styled-components": "^5.2.1",
--- a/src/components/desktop/DesktopLayout.js
+++ b/src/components/desktop/DesktopLayout.js
@@ -1,6 +1,6 @@
 import { WindowSize } from '@reach/window-size'
 import { Route, Switch } from 'react-router-dom'
-import SplitPane from 'react-split-pane'
+import { Pane, SplitPane } from 'react-split-pane'
 import styled from 'styled-components/macro'
 
 import aboutRoutes from '../about/aboutRoutes'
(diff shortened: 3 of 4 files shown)


wbazant:
Thanks for the fix! I will try it out on Monday. Why was the change in craco.config.js necessary?

wbazant:
Also, could you post a screenshot? We want two vertical panes.

nopaalh:
The react-split-pane package ships ESM build that imports react/jsx-runtime without a file extension and webpack 5 enforces strict ES Modules resolution. When it detects a module is fully specified ESM, import must include its exact file extension. The package's import's statement omits the extension, so webpack throw 'Module not found : cant resolve react/jsx-runtime' even though the file exist there in node_module

And here is the screenshot
image

nopaalh:
Should I switch to making webpack treat these as not fully specified (via fullySpecified: false in the module rules), instead of the manual alias? That should avoid triggering CRA's ModuleScopePlugin since it won't resolve to an absolute path outside src/.

wbazant:
Thanks, looks like it works! Our v2 also has a resize feature where you can drag the bar and change size but I guess that works too. I understand your suggestion - telling webpack to not assume a fully correct setup for this specific package since it isn't by webpack standards, but works anyway? - I think it's the right intention, if you can get `yarn start` and `yarn build` to succeed without warnings we'll be ready to merge.

nopaalh:
Confirmed. Both `yarn start` and `yarn build` complete successfully after switching to `fullySpecified: false`. We still can change the size of the panel, but the size of resizer area is very small. i haven't change the area. wait lemme fix it

Minimize: 
Screenshot 2026-09-13 153800

Maximize:
Screenshot 2026-09-13 153746

ezwelty:
@nopaalh Thanks for this!

So remarkably, the new version doesn't seem to support right-to-left (RTL) text out of the box – switch to Hebrew or Arabic to see what I mean. Maybe helpful: https://github.com/tomkp/react-split-pane/pull/262

nopaalh:
okayy ill check it after class today

nopaalh:
Thanks for the context! Just want to confirm a couple things before I dig further:

1. By "out of the box" you mean it doesn't work with the default settings (no extra config/props needed), right?
2. Could you point to which specific feature isn't RTL-friendly? I want to confirm if it's specifically the resize drag logic (dragging feels inverted, moving right grows instead of shrink, or vice versa) rather than something else like text alignment or layout mirroring.

I checked the current production version (before this PR's changes) and the resize logic seems to have the same issue there too, so this looks like a pre-existing limitation in react-split-pane itself, not something introduced by the v2 to v3 migration. Want to make sure I'm scoping this correctly before deciding whether to fix it here or track it separately.

ezwelty:
@nopaalh 

1. Yes, as is. Didn't want to rule out that some configuration or wrapping in an explicit `<... direction="rtl">` might be possible.
2. Yes exactly, the drag logic is in the wrong direction. Eek, I could swear it used to work (we thoroughly tested when we added RTL languages), but you're right, the same issue is present on the live site!

The patch would be minor:
https://github.com/tomkp/react-split-pane/pull/262/changes#diff-08bcb2b5339aba9db1cd795724e2615cfca754042c68e2d776e267b7944f6aab

Could be a use case for `yarn patch`, but this would require updating to yarn 2+.

wbazant:
That's a really cool bug with the resize not working in the right direction, I've never noticed it. I've merged because the issue is completed, but we have a follow up issue. @nopaalh feel free to pick it up if you like, or I can have a go at some point.

Since we don't need like 90% of the features from react-split-pane, maybe there's a forked version that gets this right, or we could vibe code a replacement.

PR #1209, opened 2026-09-13
#1198 Fix image thumbnail aspect ratio

### LINKED ISSUE #1198: Make image thumbnails 100x100, with black backgrounds
Follow-up from https://github.com/falling-fruit/falling-fruit-web/pull/1192 . Now a 70x100 px image produces a 70x70 square.

Fit to the outer edge instead, so the thumbnail looks like the full image it opens to.

MRVx77:
Closes #1198 

The thumbnails were forced into a square because the container had a fixed height and the image used object-fit: cover, causing non-square images to be cropped.

Updated the thumbnail to use height: auto and object-fit: contain, allowing images to preserve their original aspect ratio and display fully without being forced into a square.


wbazant:
Thanks, looks like good progress! I added two apple pictures to location 2292300 for a test case:

Screenshot From 2026-09-14 10-37-30

Let's fix the white background and the rectangular shape - In the reviews tab I would like squares with black blackground that show the whole image instead of crop to the centre, so the preview matches what opens in the gallery.

Also, we can add screenshots of different shapes for testing the review form. There they actually don't need to be square, but the heights should match. Adding a tall thin rectangle does this:

Screenshot From 2026-09-14 10-43-47

MRVx77:
@wbazant it done i have remove the white gaps, and also made the review images into 48x48 square. Also fix the upload image for new locations. here are the images
Screenshot

Screenshot

wbazant:
Hi, thanks for additional work on this. I think it's not the direction I want to take the solution in, sorry - my goal here is to make the thumbnail previews look like the images they open to later. I also like squares as layout, there's no need to replace them with rectangles. I would like each image as:
- a square
- with black bits of background to make up for the image source not being a square
- the image entirely shown
- same size as currently on the photo upload, and bumped up to 100px in the review tab
Let's close this PR, but feel free to open another one.

PR #1212, opened 2026-09-14
Flex the whole top bar so the type tree can grow

### LINKED ISSUE #448: fix max height of opened filter pane to just over the bottom menu icons
#416 proposes:
- Make the filter pane fullscreen (extend to footer)

I tried it, and it seemed a bit wasteful when the map is zoomed in and just a few locations are available. The proposed resolution is to flex the tree select, and let it be as big as it wants to be, with some bounds.

fa96e6b20f455546b2ae79be6dbe8443b81f0b39 adds 5vh as the minimum (a tiny bit of white space so the user knows the feature is still there) and 65vh as max ( a tiny bit of map still visible so the user knows tha map is still there).

wbazant:
Closes #448


wbazant:
@ezwelty what do you think? Stretch to the menu buttons, or to the safe area at the bottom of the screen?

ezwelty:
@wbazant Hmm, I think to the bottom? There's no reason no access the menu buttons while having the filter open, and having them hidden allows me to focus (and have more space for) composing my filter.

Issue #1144, opened 2026-07-15
Improve the drawer


wbazant:
Reported from iOS but probably any real device can be used here.

Location pull up bar stutters after it’s clicked instead of dragged:

https://github.com/user-attachments/assets/4158eeaa-6750-403c-98eb-8383815b9753

Location pull up bar too sticky:

https://github.com/user-attachments/assets/63eb30d2-3243-4112-93d8-4e1589ec7599

For the second one, we could just move the boundary between reverting and going into full pane down by a little bit. Fancy implementation would take into account acceleration of dragging but probably hard in a webpage?

wbazant:
Not quite the right size of the preview image now

https://github.com/user-attachments/assets/b713bca9-9dc8-4d8d-89e1-7f7cc9d9570a

PR #1215, opened 2026-09-14
Take out the window size resizer

### LINKED ISSUE #1093: glitch when url bar meets drawer
Mobile firefox.

https://github.com/user-attachments/assets/e34f24cc-eecf-46fc-96c2-3f0d2de21f4a

### LINKED ISSUE #1211: Support variable vertical height
In mobile browsers, height of the viewport can change - hiding URL bar on firefox mobile but most commonly the floating keyboard. We do okay, but we often don't handle things smoothly.

1) map doesn't respond to hiding the address bar on mobile firefox - on mobile, if you're on map page, and drag up from the map button, a white rectangle appears under the map.

2) The location drawer responds poorly if you start place search (we could just dismiss the drawer then)

3) The clever newly added https://github.com/falling-fruit/falling-fruit-web/pull/1212 type layout doesn't resize when starting to type, so suddenly you're able to scroll outside the container

4) on Firefox, the URL bar appearing makes the drawer go up with animation rather than without: https://github.com/falling-fruit/falling-fruit-web/issues/1093

wbazant:
map absolute -> fixed means no white url bar.

Allow some pages to have the bar going upwards, sadly with /about/welcome this triggers a glitchiness as url tries to go up simultaneously with tab list getting painted.

Closes #1211 
Closes #1093


wbazant:
@ezwelty I'll optimistically merge this one, but FYI! I think the issues I was experiencing are limited to mobile Firefox, my mobile Chrome doesn't try to cleverly hide the URL bar on scroll down - the app is also chrome based.

wbazant:
https://github.com/falling-fruit/falling-fruit-web/issues/1144#issuecomment-5668607340 found this, i think related

PR #1229, opened 2026-09-16
Drawer improvements

### LINKED ISSUE #827: Disable the carousel swipe for partially open drawer
We allow browsing between the photos, but don't let the user know it's possible, and already offer two actions there:
- click to open photo
- swipe up to open drawer
 
So the user will probably do it by accident, and be confused.

We tried to tell the user it's possible to swipe with sideways dots (#741) but it doesn't seem that good. Let's disable the swipe sideways when the drawer is partially open - this is a property in redux:
```
const drawerFullyOpen = useSelector(
    (state) => state.location.pane.drawerFullyOpen,
)
```

Not sure whether we should also disable the carousel moving left and right by itself? Clicking to open photo, or swiping up to open drawer, should still work.

### LINKED ISSUE #1091: Undesirable competing scrolls in drawer and content
When we scroll up, we open a drawer, and we also scroll inside. Not sure if I want this.

See https://github.com/user-attachments/assets/66cda826-e4b3-4c3b-a72c-2f86efb2b26d , from #1072.

### LINKED ISSUE #1144: Improve the drawer

### LINKED ISSUE #1219: loading a location with image from activity page shows a bit of map
This location for example: 
https://fallingfruit.org/locations/2480160/-17.5587848,-52.5545642,18z?pane=full

> 
> https://github.com/user-attachments/assets/b713bca9-9dc8-4d8d-89e1-7f7cc9d9570a 

 _Originally posted by @wbazant in [#1144](https://github.com/falling-fruit/falling-fruit-web/issues/1144#issuecomment-5668607340)_

### LINKED ISSUE #1224: Report modal persists after close
This is a regression - fallingfruit.org doesn't have the problem, but falling-fruit-web.pages.dev does. Try to report a location, then change your mind. Too bad - ?report=true stays in the URL, so any location you're going to open will have the modal.

wbazant:
Closes #827 
Closes #1091
Closes #1144
Closes #1224
Closes #1219

I changed the implementation to support two code paths rendering the location content: a sheet where you peek at it, and a full page. Places that need to open a full page (e.g. going to the drawer from the activity page) can now draw a top to bottom layout which rendered weird in the past. This required moving some state to redux because it now needs to be synced between two copies of the location content, for example the carousel.

It helped me make progress on other drawer bugs:
- I changed the mechanics of the drawer a bit to incorporate move velocity so a quick flick up pushes it to the top
- don't scroll inside the drawer when viewing it as a sheet


wbazant:
@ezwelty you can just be on the loop for this one as the behaviour changes are all trivial but FYI big code change!

Somehow it felt great to bring out the big guns - I had a harness drawing pages side by side and calculating pixel diffs, and everything. The actual benefit from the big guns having been brought out is just the render of the full pane (e.g. from activity page) not getting a glitch as a picture loads, and I could have done everything else with smaller changes.

